PR #223, opened 2022-02-14
fix(logrocket): workaround for stitches using insertRule

triptu:
### Details(context, Jira ticket, how was the bug fixed, what does the new feature do)

Root cause - LogRocket doesn't behave nicely with CSSStyleSheet's insert rule, this change hijacks insert rule to add styles in dom as well so Mutation Observer can observe them. The style sync is debounced for performance so there would be at max a delay of 1 seconds between broken styles and fixed styles when viewing logrocket.

- https://github.com/modulz/stitches/issues/873
- https://gist.github.com/oeduardoal/499923b72422e4222c5073ba2a708ad1
- https://github.com/styled-components/styled-components/issues/1603
- improve logrocket indetification, add session id and room id
- make logrocket url available on window - `window.logrocketURL` can be used to get logrocket url now
- update logrocket

### Choose one of these(put a 'x' in the bracket):

- [ ] The change doesn't require a change to the documentation.
- [ ] The documentation is updated accordingly.

### DIFF
--- a/apps/100ms-custom-app/src/index.js
+++ b/apps/100ms-custom-app/src/index.js
@@ -8,17 +8,8 @@ import './index.css';
 import '100ms_edtech_template/dist/index.css';
 
 if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_LOGROCKET_ID) {
-  /**
-   * TODO: starting logrocket in a timeout because it's not picking up the stitches styles otherwise for
-   * some reason, and the recording appears broken. The timeout appears to be somehow solving the issue,
-   * more details will be need to looked into to figure out the exact cause. It's likely that the stitches library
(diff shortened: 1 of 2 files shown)


harrisonhenri:
Hey guys, this solution is based on https://github.com/stitchesjs/stitches/issues/873 isn't is? Is anyone experiencing problems with this workaround when the user connection is slow?

triptu:
> Hey @raviteja83 and @triptu, this solution is based on [stitchesjs/stitches#873](https://github.com/stitchesjs/stitches/issues/873) isn't it? Is anyone experiencing problems with this workaround when the user connection is slow?

It shouldn't increase the bundle size though there will be small performance overhead at runtime. Can you give more details? The performance tab in devtools should help in figuring out the root cause.

PR #2439, opened 2024-01-18
feat: add duration metric to responses, leaderboard and participation summary

KaustubhKumar05:
image

image

image

### DIFF
--- a/packages/roomkit-react/src/Button/Button.tsx
+++ b/packages/roomkit-react/src/Button/Button.tsx
@@ -143,8 +143,8 @@ const StyledButton = styled('button', {
         '$secondary_bright',
         '$secondary_dim',
         '$secondary_disabled',
-        '$on_surface_high',
-        '$on_surface_low',
+        '$on_secondary_high',
+        '$on_secondary_low',
       ),
       danger: getButtonVariants(
         '$alert_error_default',
--- a/packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
@@ -1,5 +1,5 @@
 // @ts-check
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useRef, useState } from 'react';
 import { selectLocalPeer, selectLocalPeerRoleName, useHMSActions, useHMSStore } from '@100mslive/react-sdk';
 import { CheckCircleIcon, ChevronDownIcon, CrossCircleIcon } from '@100mslive/react-icons';
 import { Box, Button, Flex, Text } from '../../../../';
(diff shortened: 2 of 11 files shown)


packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
```diff
@@ -48,27 +41,13 @@ export const QuestionCard = ({
 
   const isLive = pollState === 'started';
   const canRespond = isLive && !localPeerResponse;
-
+  const startTime = Date.now();
```
raviteja83, line 44: no chance of re-render here? probably should be in a ref.

packages/roomkit-react/src/Button/Button.tsx
```diff
@@ -143,8 +143,8 @@ const StyledButton = styled('button', {
         '$secondary_bright',
         '$secondary_dim',
         '$secondary_disabled',
-        '$on_surface_high',
-        '$on_surface_low',
+        '$on_secondary_high',
```
raviteja83, line 146: this fixes the start recording button as well?

KaustubhKumar05: The tokens are now correct but proper color selection is also needed on dashboard by the user

packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
KaustubhKumar05: Updated and tested

packages/roomkit-react/src/Button/Button.tsx
raviteja83: the color selection should be fine. as we show the preview there, we can put it on the user to be more mindful

packages/roomkit-react/src/Prebuilt/components/Polls/Voting/TimedVoting.tsx
```diff
@@ -9,29 +9,31 @@ export const TimedView = ({ poll }: { poll: HMSPoll }) => {
...
-        answer={question.answer}
-        setCurrentIndex={setCurrentIndex}
-        rolesThatCanViewResponses={poll.rolesThatCanViewResponses}
-      />
-    ) : (
-      <></>
-    ),
+  return (
+    <>
+      {poll.questions?.map(question =>
```
raviteja83, line 17: why not use activeQuestion itslef here?

KaustubhKumar05: Had tried that first, rendering just one QuestionCard element. The radio group selection does not reset on rerender and selection from previous question remains. Tried the HTML input as well, this seems to be the expected behaviour

raviteja83 (approved):

PR #2440, opened 2024-01-18
fix: hide anonymous toggle

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Polls/CreatePollQuiz/PollsQuizMenu.jsx
```diff
@@ -141,12 +140,12 @@ const AddMenu = () => {
             Hide Vote Count
           </Text>
         </Flex>
-        <Flex align="center" css={{ mt: '$10' }}>
+        {/* <Flex align="center" css={{ mt: '$10' }}>
```
raviteja83, line 143: is this component same for polls and quizzes or are we just removing for quiz?

raviteja83 (approved):

PR #2442, opened 2024-01-18
fix: ui freeze on mweb in bring on stage flow

eswarclynn:

-
-


packages/roomkit-react/src/Prebuilt/components/Notifications/PermissionErrorModal.tsx
```diff
@@ -44,7 +44,14 @@ export function PermissionErrorModal() {
     <Dialog.Root open={!!deviceType}>
       <Dialog.Portal>
         <Dialog.Overlay />
-        <Dialog.Content css={{ w: 'min(380px, 90%)', p: '$8' }}>
+        <Dialog.Content
+          css={{
+            w: 'min(380px, 90%)',
+            p: '$8',
+            // overlay over Sheet.tsx
+            zIndex: 1003,
```
raviteja83, line 52: why does sheet have z-index in 1000? let's reduce the sheet index as well, somewhere in 10-20 range

raviteja83 (approved):

PR #2444, opened 2024-01-19
fix: process plugins on device change

KaustubhKumar05:

- Currently changing the video input results in black tile if virtual background plugins are applied since the plugin has the older stream ID
- The fix processes the plugins when the device changes
- Tested on OBS virtual camera, to test on a real device

### DIFF
--- a/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
+++ b/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
@@ -182,6 +182,9 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       // if track is muted, we just cache the settings for when it is unmuted
       this.settings = newSettings;
       return;
+    } else {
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -182,6 +182,9 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       // if track is muted, we just cache the settings for when it is unmuted
       this.settings = newSettings;
       return;
+    } else {
+      await this.pluginsManager.waitForRestart();
+      this.processPlugins();
```
raviteja83, line 187: await?

KaustubhKumar05: Yup, returns a promise

raviteja83 (approved):

PR #2460, opened 2024-01-23
fix: black tile when vb is enabled in preview

raviteja83:

-
-

### DIFF
--- a/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
+++ b/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
@@ -184,7 +184,6 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       return;
     } else {
       await this.pluginsManager.waitForRestart();
-      await this.processPlugins();
     }
     await this.handleSettingsChange(newSettings);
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -184,7 +184,6 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       return;
     } else {
       await this.pluginsManager.waitForRestart();
```
KaustubhKumar05, line 186: Shouldn't we call waitForRestart along with processPlugins? Or are we deprecating HMS VB?

raviteja83: the way both of them is different, hence it should not cause issue for previous plugin

KaustubhKumar05: Tested on HMS VB with the same scenarios, working fine
- Video toggle
- Device change
- Joining with VB

KaustubhKumar05 (approved):

PR #2462, opened 2024-01-24
fix: scroll only for too many options

eswarclynn:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
@@ -135,7 +135,9 @@ export const QuestionCard = ({
         </Box>
       </Flex>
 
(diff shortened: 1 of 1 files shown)


KaustubhKumar05:
image

The dropdown icon animates but the content is not hidden. Also can we just show the entire content without scroll for parity with other platforms in expanded view?

packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
```diff
@@ -135,52 +135,50 @@ export const QuestionCard = ({
         </Box>
       </Flex>
 
-      <Box css={{ maxHeight: showOptions ? '$80' : '0', transition: 'max-height 0.3s ease', overflowY: 'hidden' }}>
```
KaustubhKumar05, line 138: The showOptions state is not being used

raviteja83 (approved):

PR #2464, opened 2024-01-24
fix: quiz duration formatting

KaustubhKumar05:

- Format time into minutes and seconds, fix decimal places to match designs

image

image

image

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Polls/common/utils.ts
+++ b/packages/roomkit-react/src/Prebuilt/components/Polls/common/utils.ts
@@ -1,12 +1,16 @@
 export const getFormattedTime = (milliseconds: number | undefined) => {
   if (!milliseconds) return '-';
+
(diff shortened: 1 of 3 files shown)


packages/roomkit-react/src/Prebuilt/components/Polls/common/utils.ts
```diff
@@ -0,0 +1,12 @@
+export const getFormattedTime = (milliseconds: number | undefined) => {
+  if (!milliseconds) return '-';
+  const totalSeconds = milliseconds / 1000;
+  const minutes = Math.floor(totalSeconds / 60);
+  const seconds = totalSeconds % 60;
+
+  const timeString = `${minutes ? `${minutes}m ` : ''}${
```
raviteja83, line 7: use if cond to be more cleaner.

raviteja83 (approved):

PR #2470, opened 2024-01-25
fix: handle reopening quiz after partial attempt

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Polls/Voting/PeerParticipationSummary.tsx
```diff
@@ -33,12 +33,12 @@ export const PeerParticipationSummary = ({ quiz }: { quiz: HMSPoll }) => {
         { title: 'Avg. Time Taken', value: getFormattedTime(summary.avgTime) },
         {
           title: 'Avg. Score',
-          value: Number.isInteger(summary.avgScore) ? summary.avgScore : summary.avgScore.toFixed(2),
+          value: Number.isInteger(summary.avgScore) ? summary.avgScore : summary.avgScore.toFixed(2) || 0,
```
raviteja83, line 36: why is there no `?.` if we are doing || ?

KaustubhKumar05: summary is defined, avgscore can be undefined

KaustubhKumar05: This is just needed for the score, will remove for avgScore and check

packages/roomkit-react/src/Prebuilt/components/Polls/common/SingleChoiceOptions.jsx
```diff
@@ -17,10 +17,9 @@ export const SingleChoiceOptions = ({
   isStopped,
   isQuiz,
   localPeerResponse,
-  answer,
 }) => {
   return (
-    <RadioGroup.Root value={answer || null} onValueChange={value => setAnswer(value)}>
+    <RadioGroup.Root value={localPeerResponse?.option} onValueChange={value => setAnswer(value)}>
```
raviteja83, line 22: shouldn't this be || null?

KaustubhKumar05: Not requried now since multiple cards need to rendered

raviteja83 (approved):

PR #2475, opened 2024-01-29
fix: add manual device change analytics

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/device-manager/DeviceManager.ts
+++ b/packages/hms-video-store/src/device-manager/DeviceManager.ts
@@ -49,6 +49,20 @@ export class DeviceManager implements HMSDeviceManager {
         this.eventBus.deviceChange.publish({ devices: this.getDevices() } as HMSDeviceChangeEvent);
       }
     });
+
+    this.eventBus.manualDeviceChange.subscribe(({ type, deviceId, groupId }) => {
+      const inputType = type === 'video' ? 'videoInput' : type;
+      const newSelection = this[inputType].find(
+        device => this.createIdentifier(device) === this.createIdentifier({ deviceId, groupId }),
+      );
+      this.eventBus.analytics.publish(
+        AnalyticsEventFactory.deviceChange({
+          selection: { [inputType]: newSelection },
+          devices: this.getDevices(),
+          type,
+        }),
+      );
+    });
   }
 
   updateOutputDevice = async (deviceId?: string) => {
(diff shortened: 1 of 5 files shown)


packages/hms-video-store/src/events/EventBus.ts
```diff
@@ -18,6 +18,9 @@ import { ITrackAudioLevelUpdate } from '../utils/track-audio-level-monitor';
 export class EventBus {
   private eventEmitter: EventEmitter = new EventEmitter();
   readonly deviceChange = new HMSInternalEvent<HMSDeviceChangeEvent>(HMSEvents.DEVICE_CHANGE, this.eventEmitter);
+  readonly manualDeviceChange = new HMSInternalEvent<
```
raviteja83, line 21: we could have used the device change and added a property to manualDeviceChange no?

eswarclynn: The problem with that is it has a devices property which specifies all the devices present and we don't get this in the track class when it requires to be published.

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -136,19 +136,21 @@ export default class AnalyticsEventFactory {
   }
 
   static deviceChange({
+    isManual,
     selection,
     type,
     devices,
     error,
   }: {
+    isManual?: boolean;
```
raviteja83, line 145: userSelection instead of manual

packages/hms-video-store/src/device-manager/DeviceManager.ts
```diff
@@ -49,6 +49,23 @@ export class DeviceManager implements HMSDeviceManager {
         this.eventBus.deviceChange.publish({ devices: this.getDevices() } as HMSDeviceChangeEvent);
       }
     });
+
+    this.eventBus.deviceChange.subscribe(({ type, isManual, selection }) => {
+      if (isManual) {
```
raviteja83, line 54: can we not do this in the handleDeviceChange in sdk/index? there we have access to devicemanager and can get the devices

packages/hms-video-store/src/utils/constants.ts
```diff
@@ -36,6 +36,7 @@ export const SUBSCRIBE_STATS_PUSH_INTERVAL = 60;
 
 export const HMSEvents = {
   DEVICE_CHANGE: 'device-change',
+  MANUAL_DEVICE_CHANGE: 'manual-device-change',
```
raviteja83, line 39: no longer needed

packages/hms-video-store/src/device-manager/DeviceManager.ts
eswarclynn: problem is getDevices method of DeviceManager is not public. No problem if we make it public?

eswarclynn: createIdentifier too

raviteja83 (approved):

PR #2492, opened 2024-02-05
feat: default vb set by layout

KaustubhKumar05:
### Details

- Default VB media should be auto enabled when user joins preview


packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBToggle.tsx
```diff
@@ -1,15 +1,24 @@
...
 
 export const VBToggle = () => {
   const toggleVB = useSidepaneToggle(SIDE_PANE_OPTIONS.VB);
   const isVBOpen = useIsSidepaneTypeOpen(SIDE_PANE_OPTIONS.VB);
   const isVideoOn = useHMSStore(selectIsLocalVideoEnabled);
+  const { elements } = useRoomLayoutConferencingScreen();
+  const backgroundMedia = elements?.virtual_background?.background_media;
+
+  useEffect(() => {
+    // If default is present, open the VB sidepane
```
KaustubhKumar05, line 19: Will use the same sessionStorage key here with another value - `opened` and `set` to prevent the sidepane from opening and setting the VB multiple times

packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -122,14 +123,14 @@ export const HMSPrebuilt = React.forwardRef<HMSPrebuiltRefType, HMSPrebuiltProps
       (ref as MutableRefObject<HMSPrebuiltRefType>).current = { ...reactiveStore.current };
     }, [ref]);
 
-    // leave room when component unmounts
-    useEffect(
-      () => () => {
+    useEffect(() => {
+      sessionStorage.setItem(DEFAULT_VB_STATE, '');
```
raviteja83, line 127: why do we need sessionStorage?

raviteja83: to check whether vb is selected or not?

KaustubhKumar05: Can add another key in appData - started implementation with this since it's faster

KaustubhKumar05: Wanted to get a version out to get UX feedback

raviteja83: yes, we can add. we can set in appData initially when we load the layout response

packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBPicker.tsx
```diff
@@ -129,16 +146,16 @@ export const VBPicker = ({ backgroundMedia = [] }: { backgroundMedia: VirtualBac
               value: HMSVirtualBackgroundTypes.BLUR,
               onClick: async () => {
                 await VBHandler?.setBlur(blurAmount);
-                setActiveBackground(HMSVirtualBackgroundTypes.BLUR);
+                hmsActions.setAppData(APP_DATA.background, HMSVirtualBackgroundTypes.BLUR);
```
raviteja83, line 149: so the backgroundType in appData is also not needed?

KaustubhKumar05: Nope, removed it now

raviteja83 (approved):

PR #2496, opened 2024-02-06
fix: improve chat UX for small heights

KaustubhKumar05:
### Details

- The empty chat image is hidden for small screens https://100ms.atlassian.net/browse/WEB-2612

image

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Chat/EmptyChat.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Chat/EmptyChat.tsx
@@ -33,13 +33,12 @@ export const EmptyChat = () => {
       justify="center"
     >
       <Box>
-        {window && window.screen.height > 600 ? (
-          <img src={emptyChat} alt="Empty Chat" height={132} width={185} style={{ margin: '0 auto' }} />
-        ) : null}
+        <img src={emptyChat} alt="Empty Chat" height={132} width={185} style={{ margin: '0 auto' }} />
+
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/components/Chat/EmptyChat.tsx
```diff
@@ -33,11 +33,13 @@ export const EmptyChat = () => {
       justify="center"
     >
       <Box>
-        <img src={emptyChat} alt="Empty Chat" height={132} width={185} style={{ margin: '0 auto' }} />
+        {window && window.screen.height > 600 ? (
```
raviteja83, line 36: better to add scroll to this section

packages/roomkit-react/src/Prebuilt/components/Chat/EmptyChat.tsx
```diff
@@ -33,11 +33,13 @@ export const EmptyChat = () => {
...
+          <img src={emptyChat} alt="Empty Chat" height={132} width={185} style={{ margin: '0 auto' }} />
+        ) : null}
         <Text variant="h5" css={{ mt: '$8', c: '$on_surface_high' }}>
           {canSendMessages ? 'Start a conversation' : 'No messages yet'}
         </Text>
-        {canSendMessages ? (
+        {canSendMessages && window.screen.height > 400 ? (
```
raviteja83, line 42: we should not be doing it like this

raviteja83 (approved):

PR #2501, opened 2024-02-06
feat: show chat notifications when screensharing and call tab is not active

eswarclynn:

- permission is requested after screen share is successfully published
- notification is shown when screen is shared and the video call tab is not active 

### Implementation note, gotchas, related work and Future TODOs (optional)

Compatibility notes: https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification#chrome_notes


packages/react-sdk/src/hooks/useAwayNotifications.ts
```diff
@@ -0,0 +1,28 @@
...
+export const useAwayNotifications = () => {
+  const requestPermission = useCallback(async () => {
+    if (!Notification || Notification?.permission === 'granted') {
+      return;
+    }
+    await Notification.requestPermission();
+  }, []);
+
+  const showNotification = useCallback((title: string, options?: NotificationOptions) => {
+    if (!Notification || Notification?.permission === 'denied' || document.visibilityState === 'visible') {
```
raviteja83, line 13: shouldn't we do request permission internally here also when denied?

raviteja83: I think we should also have some UI saying that you have to enabled from settings at system level if not allowed already. Granted might be true but if not enabled at system level, it will not be shown.

packages/roomkit-react/src/Prebuilt/components/Notifications/Notifications.tsx
```diff
@@ -182,6 +187,14 @@ export function Notifications() {
           delete pollToastKey[notification?.data.id];
         }
         break;
+      case HMSNotificationTypes.NEW_MESSAGE:
```
raviteja83, line 190: we have a separate chatNotifications file

PR #2503, opened 2024-02-06
fix: remove sidebar mode for mweb

raviteja83:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/VideoLayouts/ScreenshareLayout.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/VideoLayouts/ScreenshareLayout.tsx
@@ -35,10 +35,19 @@ export const ScreenshareLayout = ({ peers, onPageChange, onPageSize, edgeToEdge 
   }, [activeSharePeer, peers, isMobile, layoutMode]);
 
   useEffect(() => {
-    if (layoutMode !== LayoutMode.SIDEBAR) {
-      setLayoutMode(LayoutMode.SIDEBAR);
+    if (isMobile) {
+      setLayoutMode(LayoutMode.GALLERY);
+      return;
     }
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+    if (layoutMode === LayoutMode.SIDEBAR) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/components/VideoLayouts/ScreenshareLayout.tsx
```diff
@@ -35,10 +35,19 @@ export const ScreenshareLayout = ({ peers, onPageChange, onPageSize, edgeToEdge
   }, [activeSharePeer, peers, isMobile, layoutMode]);
 
   useEffect(() => {
-    if (layoutMode !== LayoutMode.SIDEBAR) {
-      setLayoutMode(LayoutMode.SIDEBAR);
```
KaustubhKumar05, line 39: It worked last time with the hasSidebar prop being set to false. Do we still need that?

packages/roomkit-react/src/Prebuilt/components/VideoLayouts/ScreenshareLayout.tsx
```diff
@@ -35,10 +35,19 @@ export const ScreenshareLayout = ({ peers, onPageChange, onPageSize, edgeToEdge
...
+    if (isMobile) {
+      setLayoutMode(LayoutMode.GALLERY);
+      return;
     }
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+    if (layoutMode === LayoutMode.SIDEBAR) {
+      return;
+    }
```
KaustubhKumar05, line 44: Is this needed? It probably won't rerender if prevState === new value

raviteja83: it is for the cleanup to not run

packages/roomkit-react/src/Prebuilt/components/VideoLayouts/ScreenshareLayout.tsx
raviteja83: on resize also it has to be handled hence thjs

KaustubhKumar05 (approved):

PR #2505, opened 2024-02-08
fix: toggle video for mweb when in background

KaustubhKumar05:


packages/hms-video-store/src/utils/user-agent.ts
```diff
@@ -77,3 +77,10 @@ const convertObjectToString = (object: UserAgent, delimiter = ',') =>
     .filter(key => isPresent(object[key as keyof UserAgent]))
     .map(key => `${key}:${object[key as keyof UserAgent]}`)
     .join(delimiter);
+
+export const isMobile = () => {
+  if (window && window.screen.width < 768) {
```
raviteja83, line 82: should we use userAgent instead?

KaustubhKumar05:
Closed in favour of https://github.com/100mslive/web-sdks/pull/2571/files

PR #2527, opened 2024-02-15
feat: Room title and description

KaustubhKumar05:
Pending:
- [x] Design for the scenario when only the description is present. No way to open the sheet in mweb if the title is missing
- [x] To remove dummy values added for testing before alpha publish


packages/roomkit-react/src/Prebuilt/layouts/SidePane.tsx
```diff
@@ -52,6 +53,9 @@ const SidePane = ({
   if (sidepane === SIDE_PANE_OPTIONS.VB) {
     ViewComponent = <VBPicker backgroundMedia={backgroundMedia} />;
   }
+  if (sidepane === SIDE_PANE_OPTIONS.ROOM_DETAILS) {
```
raviteja83, line 56: do explore what we can do here. this is gonna explode someday. we should start looking for alternative approaches here

packages/roomkit-react/src/Prebuilt/provider/roomLayoutProvider/hooks/useRoomLayoutScreen.ts
```diff
@@ -69,3 +79,38 @@ export function useRoomLayoutLeaveScreen() {
...
+  const { isRecordingOn } = useRecordingStreaming();
+  const peerCount = useHMSStore(selectPeerCount);
+  const sessionStartedAt = useHMSStore(selectRoomStartTime);
+  const hlsState = useHMSStore(selectHLSState);
+
+  // People watching, start timestamp, recording status
+  const details = useMemo(() => {
+    const details = [];
+    let timeStamp;
+    if (hlsState?.variants?.[0] && hlsState?.variants[0]?.['startedAt']) {
```
raviteja83, line 95: second condition is enough no?

packages/roomkit-react/src/Prebuilt/provider/roomLayoutProvider/hooks/useRoomLayoutScreen.ts
```diff
@@ -69,3 +79,38 @@ export function useRoomLayoutLeaveScreen() {
...
+      details.push(`${getFormattedCount(peerCount)} watching`);
+      const streamStartedAt = hlsState?.variants[0]['startedAt'];
```
raviteja83, line 97: this var seems unnecessary

packages/roomkit-react/src/Prebuilt/provider/roomLayoutProvider/hooks/useRoomLayoutScreen.ts
```diff
@@ -69,3 +79,38 @@ export function useRoomLayoutLeaveScreen() {
...
+      details.push(timeStamp);
+    }
+    if (isRecordingOn) {
+      details.push('Recording');
+    }
+    return details;
+  }, [hlsState?.variants, isRecordingOn, peerCount, sessionStartedAt]);
+
+  return {
+    title: elements.header?.title || 'Title',
```
raviteja83, line 112: should we simply hide if no title? showing Title or Description when empty is meaningless

packages/roomkit-react/src/Prebuilt/provider/roomLayoutProvider/hooks/useRoomLayoutScreen.ts
raviteja83: `?.` not needed here

packages/roomkit-react/src/Prebuilt/components/RoomDetails/Duration.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { useEffect, useState } from 'react';
+import { Flex } from '../../../Layout';
+import { Text } from '../../../Text';
+import { getFormattedTime } from '../Polls/common/utils';
+
+export const Duration = ({ timeStamp }: { timeStamp: Date }) => {
```
raviteja83, line 6: no need of camelCase for timestamp

packages/roomkit-react/src/Prebuilt/components/Polls/common/utils.ts
```diff
@@ -1,16 +1,22 @@
...
-    formattedSeconds = seconds.toFixed(1);
+  let formattedTime = '';
+  if (hours) {
+    formattedTime += `${hours}h `;
   }
+  if (minutes || hours) {
+    formattedTime += `${minutes}m `;
+  }
+  if (!precise && (hours || minutes)) {
+    return formattedTime;
```
raviteja83, line 17: seconds will be missing here. does it break the previous usage anywhere?

KaustubhKumar05: This in intentional. The default value of precise is true, which is used everywhere else

packages/roomkit-react/src/Prebuilt/provider/roomLayoutProvider/hooks/useRoomLayoutScreen.ts
KaustubhKumar05: Updated

packages/roomkit-react/src/Prebuilt/layouts/Sheet.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { selectAppData, useHMSStore } from '@100mslive/react-sdk';
+import { RoomDetailsSheet } from '../components/RoomDetails/RoomDetailsSheet';
+import { Box } from '../../Layout';
+import { APP_DATA, SHEET_OPTIONS } from '../common/constants';
+
+export const Sheet = () => {
```
raviteja83, line 7: why do we need this separate component?

packages/roomkit-react/src/Prebuilt/components/AppData/AppData.tsx
```diff
@@ -54,6 +54,7 @@ const initialAppData = {
   },
   [APP_DATA.chatDraft]: '',
   [APP_DATA.sidePane]: '',
+  [APP_DATA.sheet]: '',
```
raviteja83, line 57: why do we need this?

packages/roomkit-react/src/Prebuilt/layouts/Sheet.tsx
KaustubhKumar05: It can be triggered via the header as well as the tile in mweb options sheet

packages/roomkit-react/src/Prebuilt/components/AppData/AppData.tsx
KaustubhKumar05: This will help move towards a common sheet component which has been discussed before instead of having multiple sheets and individual states for all

raviteja83: better way would be to portal out rather having to condiitionally check the current state. will become same as SidePane

packages/roomkit-react/src/Prebuilt/layouts/Sheet.tsx
raviteja83: existing sheet can't be used?

packages/roomkit-react/package.json
```diff
@@ -83,11 +83,11 @@
   },
   "dependencies": {
     "@100mslive/hls-player": "0.2.6",
-    "@100mslive/hms-noise-cancellation": "0.0.0-alpha.1",
```
raviteja83, line 86: this should be present

raviteja83 (approved):

PR #2535, opened 2024-02-15
fix: sort polls

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Polls/CreatePollQuiz/PollsQuizMenu.tsx
```diff
@@ -195,6 +195,9 @@ const AddMenu = () => {
 const PrevMenu = () => {
   const hmsActions = useHMSActions();
   const polls = useHMSStore(selectPolls);
+  const sortedPolls = polls
```
raviteja83, line 198: should be in a useMemo

KaustubhKumar05: Updated

raviteja83 (approved):

PR #2552, opened 2024-02-19
fix: timestamps for publisher.stats by creating all tracks/layers samples at once

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
+++ b/packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
@@ -40,15 +40,17 @@ export class SubscribeStatsAnalytics extends BaseStatsAnalytics {
 
   protected sendEvent() {
     this.eventBus.analytics.publish(AnalyticsEventFactory.subscribeStats(this.toAnalytics()));
+    super.sendEvent();
   }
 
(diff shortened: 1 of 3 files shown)


packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
```diff
@@ -57,18 +59,28 @@ export class SubscribeStatsAnalytics extends BaseStatsAnalytics {
...
       }
+      const trackAnalytics = this.trackAnalytics.get(trackID);
+      if (trackAnalytics?.shouldCreateSample()) {
+        shouldCreateSample = true;
+      }
     });
+
+    if (shouldCreateSample) {
+      this.trackAnalytics.forEach(trackAnalytic => {
+        trackAnalytic.createSample();
```
raviteja83, line 74: can createSample internally not check if shouldCreate sample?

eswarclynn: That was kind of how it was earlier https://github.com/100mslive/web-sdks/pull/2552/files#diff-2baca99bf12c52138e59a205dfc730eb3faece28b338612199acc8067b8e328dL106

We need to decide if we should create sample at the top level and if we should, we create it for all tracks. So it cannot be checked at the track level.

raviteja83 (approved):

PR #2581, opened 2024-02-22
feat: add subscriber video stats to analytics

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
+++ b/packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
@@ -49,8 +49,12 @@ export class SubscribeStatsAnalytics extends BaseStatsAnalytics {
     Object.keys(remoteTracksStats).forEach(trackID => {
       const trackStats = remoteTracksStats[trackID];
       const track = this.store.getTrackById(trackID);
+      const calculatedJitterBufferDelay =
+        trackStats.jitterBufferDelay &&
+        trackStats.jitterBufferEmittedCount &&
+        trackStats.jitterBufferDelay / trackStats.jitterBufferEmittedCount;
       if (this.trackAnalytics.has(trackID)) {
-        this.trackAnalytics.get(trackID)?.pushTempStat({ ...trackStats });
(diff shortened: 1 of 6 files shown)


packages/hms-video-store/src/utils/constants.ts
```diff
@@ -34,6 +34,8 @@ export const PUBLISH_STATS_PUSH_INTERVAL = 300;
 export const SUBSCRIBE_STATS_SAMPLE_WINDOW = 10;
 export const SUBSCRIBE_STATS_PUSH_INTERVAL = 60;
 
+export const JAVA_INTEGER_MAX_VALUE = 2147483647;
```
raviteja83, line 37: JAVA?

raviteja83: Math.pow(2,31)-1

packages/hms-video-store/src/analytics/stats/SubscribeStatsAnalytics.ts
```diff
@@ -75,6 +82,25 @@ export class SubscribeStatsAnalytics extends BaseStatsAnalytics {
...
+  private calculateAvSyncForStat(trackStats: HMSTrackStats, hmsStats: HMSWebrtcStats) {
+    if (!trackStats.peerID || !(trackStats.kind === 'video')) {
+      return;
+    }
+    const peer = this.store.getPeerById(trackStats.peerID);
+    const audioTrack = peer?.audioTrack;
+    const videoTrack = peer?.videoTrack;
+    const areBothTracksEnabled = audioTrack && videoTrack && audioTrack.enabled && videoTrack.enabled;
+    if (!areBothTracksEnabled) {
+      return JAVA_INTEGER_MAX_VALUE;
```
raviteja83, line 96: we could use Number.POSITIVE_INFINITY or Number.MAX_SAFE_INTEGER. does not work?

eswarclynn: The value needs to be same across clients, the JS value is different from the value Java uses.

raviteja83 (approved):

PR #2582, opened 2024-02-22
feat: add polling for devices when device change not available

raviteja83:

-
-


packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -288,9 +307,14 @@ export class AudioSinkManager {
       }
       const localAudioTrack = this.store.getLocalPeer()?.audioTrack;
       if (localAudioTrack && earpiece) {
+        const otherDevices = bluetoothDevice?.deviceId || wired?.deviceId || speakerPhone?.deviceId;
```
KaustubhKumar05, line 310: Can be renamed to externalDeviceID or otherDeviceID

KaustubhKumar05 (approved):

PR #2583, opened 2024-02-22
fix: duplicate toasts

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Notifications/PeerNotifications.tsx
```diff
@@ -20,7 +20,8 @@ export const PeerNotifications = () => {
   const [selectedPeer, setPeerSelector] = useSetSubscribedChatSelector(CHAT_SELECTOR.PEER);
 
   useEffect(() => {
-    if (!notification?.data) {
+    // @ts-ignore
```
raviteja83, line 23: why ts-ignore is needed? also we can combine both the conditions

packages/roomkit-react/src/Prebuilt/components/Toast/ToastBatcher.js
```diff
@@ -40,7 +41,13 @@ export const ToastBatcher = {
   },
   showToast({ notification, duration = 3000, type }) {
     try {
-      this.showToastInternal({ notification, duration, type });
+      if (!this.toastCache[notification.id]) {
```
raviteja83, line 44: i am not sure this covers all cases

KaustubhKumar05: Should we merge to QA to DF?

raviteja83: we can df on vercel itself.

raviteja83 (approved):

PR #2587, opened 2024-02-24
feat: enable mute all for roles with appropriate permissions

raviteja83:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
@@ -20,15 +20,11 @@ import { getMetadata } from '../../common/utils';
 const dropdownItemCSS = { backgroundColor: '$surface_default', gap: '$4', p: '$8' };
 const optionTextCSS = { fontWeight: '$semiBold', color: '$on_surface_high', textTransform: 'none' };
 
-export const RoleOptions = ({ roleName, peerList }: { roleName: string; peerList: HMSPeer[] }) => {
-  const [openOptions, setOpenOptions] = useState(false);
-  const permissions = useHMSStore(selectPermissions);
-  const hmsActions = useHMSActions();
-  const { elements } = useRoomLayoutConferencingScreen();
-  const { on_stage_role, off_stage_roles = [] } = (elements as DefaultConferencingScreen_Elements)?.on_stage_exp || {};
-
+const MuteUnmuteOption = ({ roleName, peerList }: { peerList: HMSPeer[]; roleName: string }) => {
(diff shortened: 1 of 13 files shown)


packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
```diff
@@ -20,27 +27,90 @@ import { getMetadata } from '../../common/utils';
...
+          {allPeersHaveAudioOn && permissions?.mute ? (
+            <Dropdown.Item css={dropdownItemCSS} onClick={() => setTrackEnabled('audio', false)}>
+              <MicOffIcon />
+              <Text variant="sm" css={optionTextCSS}>
+                Mute Audio
+              </Text>
+            </Dropdown.Item>
+          ) : null}
+
+          {!allPeersHaveAudioOn && permissions?.unmute ? (
```
KaustubhKumar05, line 71: The permissions checks here and at L62 are redundant, used at L183

packages/roomkit-react/src/Prebuilt/primitives/DialogContent.jsx
```diff
@@ -21,9 +22,9 @@ export const DialogContent = ({ Icon, title, closeable = true, children, css, ic
           <Flex justify="between">
             <Flex align="center" css={{ mb: '$1' }}>
               {Icon ? (
-                <Box css={{ mr: '$2', color: '$on_primary_high', ...iconCSS }}>
+                <Flex css={{ mr: '$2', color: '$on_primary_high', ...flexCenter, ...iconCSS }}>
```
KaustubhKumar05, line 25: Can this break usage at different places?

packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
raviteja83: they are needed because L183 is true of any of them. sometimes, only mute/unmute can be enabled. will handle those cases

packages/roomkit-react/src/Prebuilt/primitives/DialogContent.jsx
raviteja83: should not. earlier it was not aligning the icon in the center. we can test though in df. can mention in test cases

packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
KaustubhKumar05: Oh right

KaustubhKumar05 (approved):

PR #2602, opened 2024-02-27
fix: page crash for options in hand raise accordion

raviteja83:

-
-


packages/roomkit-react/src/Prebuilt/components/Footer/RoleOptions.tsx
```diff
@@ -34,17 +39,17 @@ const MuteUnmuteOption = ({ roleName, peerList }: { peerList: HMSPeer[]; roleNam
...
   peerList.forEach(peer => {
     if (peer.isLocal) {
       return;
     }
     const isAudioOn = !!peer.audioTrack && store.tracks[peer.audioTrack]?.enabled;
     const isVideoOn = !!peer.videoTrack && store.tracks[peer.videoTrack]?.enabled;
-    allPeersHaveAudioOn = allPeersHaveAudioOn && isAudioOn;
-    allPeersHaveVideoOn = allPeersHaveVideoOn && isVideoOn;
+    isAudioOnForSomePeers = isAudioOnForSomePeers || isAudioOn;
+    isVideoOnForSomePeers = isVideoOnForSomePeers || isVideoOn;
```
KaustubhKumar05, line 52: Since an early return can be useful here, we should change to peerList.some?

raviteja83: since there is a possibility fo audio being different and video being different it might not work with some.

KaustubhKumar05 (approved):

PR #2609, opened 2024-02-28
feat: hls autoplayback button mweb

amar-1995:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/HLSAutoplayBlockedPrompt.tsx
```diff
@@ -10,6 +13,33 @@ export function HLSAutoplayBlockedPrompt({
   open: boolean;
   unblockAutoPlay: () => Promise<void>;
 }) {
+  const isLandscape = useIsLandscape();
+  const isMobile = useMedia(config.media.md);
+  if ((isMobile || isLandscape) && open) {
```
raviteja83, line 18: it can be for desktop also no? only for mweb?

amar-1995: yes

raviteja83 (approved):

PR #2610, opened 2024-02-28
fix: hls video dimensions

amar-1995:

-
-


raviteja83:
this does not have all the changes most likely still breaks.

PR #2618, opened 2024-02-29
fix: seekbar getting cropped in fullscreen

amar-1995:


packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -493,7 +493,7 @@ const HLSView = () => {
                     align="start"
                     css={{
                       position: 'absolute',
-                      bottom: '0',
+                      bottom: isFullScreen ? '2%' : '0',
```
raviteja83, line 496: why 2%? to show the seekbar? instead do fullscreen on the video container no. does it work in desktop/mobile?

amar-1995: in rest of the design seekbar is at the top of controls

amar-1995: Yes, 2% is to show seekbar

raviteja83: we can use the height of the seekbar instead no? why 2%? The height won't change right. depending on dvr use thumb also for height

packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -602,7 +602,7 @@ const HLSView = () => {
       ) : (
         <Flex align="center" justify="center" direction="column" css={{ size: '100%', px: '$10' }}>
           <Flex align="center" gap="2" css={{ position: 'absolute', left: '$4', top: '$4', zIndex: 1 }}>
-            <LeaveRoom screenType={screenType} />
+            {isMobile || isLandscape ? <LeaveRoom screenType={screenType} /> : null}
```
raviteja83, line 605: how did we miss this? kind of the first thing in hls view no?

amar-1995: I was checking after starting the stream. So never saw the `stream yet to start` screen.

raviteja83 (approved):

PR #2623, opened 2024-02-29
fix: show emoji reaction when chat paused

raviteja83:

-
-


amar-1995 (dismissed):

amar-1995 (approved):

packages/roomkit-react/src/Prebuilt/components/Chat/Chat.tsx
```diff
@@ -54,6 +62,38 @@ export const Chat = () => {
           <NewMessageIndicator scrollToBottom={scrollToBottom} listRef={listRef} />
         </ChatFooter>
       ) : null}
+      {(isMobileHLSStream || isLandscapeStream) && (
```
KaustubhKumar05, line 65: Considering we're using this in multiple places, it should have been a single hook which returns isMobileHLSStream, isLandscapeStream and isStream (which is isMobileHLSStream || isLandscapeStream)

PR #2630, opened 2024-02-29
fix: remove fraction in second

amar-1995:


packages/roomkit-react/src/Prebuilt/components/Polls/common/utils.ts
```diff
@@ -16,7 +16,7 @@ export const getFormattedTime = (milliseconds: number | undefined, precise = tru
   if (!precise && (hours || minutes)) {
     return formattedTime;
   }
-  formattedTime += `${minutes >= 1 ? Math.floor(seconds) : seconds.toFixed(3)}s`;
+  formattedTime += `${Math.floor(seconds)}s`;
```
raviteja83, line 19: decimals needed in quiz. check where else this is used

KaustubhKumar05 (approved):

PR #2635, opened 2024-03-01
fix: seekbar smoothness

amar-1995:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/VideoProgress.tsx
```diff
@@ -7,42 +7,48 @@ export const VideoProgress = () => {
...
   const [videoProgress, setVideoProgress] = useState<number>(0);
   const [bufferProgress, setBufferProgress] = useState(0);
+  const [pauseProgress, setPauseProgress] = useState(false);
   const videoEl = hlsPlayer?.getVideoElement();
 
-  const onValueChange = (time: number) => {
-    hlsPlayer?.seekTo(time);
-  };
+  const setProgress = useCallback(() => {
+    if (!videoEl) {
```
raviteja83, line 14: to be safe, use pauseProgress here instead and do return

KaustubhKumar05:
Not able to see the seekbar on tapping

packages/roomkit-react/src/Prebuilt/components/HMSVideo/VideoProgress.tsx
amar-1995: `pauseProgress` is being used to stop the `timeupdate` event while we are seeking.

amar-1995:
> 
> 
> Not able to see the seekbar on tapping

It might take a sec to load

packages/roomkit-react/src/Prebuilt/components/HMSVideo/VideoProgress.tsx
raviteja83: instead of in the useEffect, use it directly here for early return. then it will cover all cases.

amar-1995: updated

raviteja83 (approved):

PR #2637, opened 2024-03-01
fix: hls layout for mweb

raviteja83:

-
-


amar-1995 (commented):
Cross icon not visible

amar-1995 (approved):

amar-1995 (approved):
LGTM

PR #2640, opened 2024-03-04
fix: disable pip

amar-1995:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/HMSVideo.jsx
```diff
@@ -46,6 +46,7 @@ export const HMSVideo = forwardRef(({ children, ...props }, videoRef) => {
         }}
         ref={videoRef}
         playsInline
+        disablePictureInPicture
```
KaustubhKumar05, line 49: It will always be disabled?

amar-1995: yes

raviteja83 (approved):

PR #2644, opened 2024-03-04
fix: spacing, controls background

amar-1995:


packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -455,49 +455,51 @@ const HLSView = () => {
...
-                        css={{
-                          visibility: isSeekEnabled ? `` : `hidden`,
-                          opacity: isSeekEnabled ? `1` : '0',
-                        }}
-                      >
-                        <ForwardArrowIcon width={32} height={32} />
-                      </HMSVideoPlayer.Seeker>
-                    </Flex>
-                  )}
+                  <Flex
```
raviteja83, line 458: this does not cover fullscreen no? does it fix when there is no chat?

raviteja83: this is only for controls no?

raviteja83: why is the DVR check removed?

raviteja83 (approved):

PR #2664, opened 2024-03-06
feat: calculate total usage for plugin

KaustubhKumar05:
- [x] calculate total usage for a plugin
- [x] Persist value until the session ends, send event before leave

Test cases:
- Disable after 5 seconds and leave ✅ 
- Leave after keeping the plugin active for 2 periods of 5 seconds each and then disabling ✅ 
- Disable after 5 seconds. Re-enable for 10s and leave without disabling ✅

### DIFF
--- a/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
+++ b/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
@@ -3,6 +3,7 @@ import { AdditionalAnalyticsProperties } from './AdditionalAnalyticsProperties';
 import AnalyticsEvent from './AnalyticsEvent';
 import { AnalyticsEventLevel } from './AnalyticsEventLevel';
 import { IAnalyticsPropertiesProvider } from './IAnalyticsPropertiesProvider';
+import { pluginUsageTracker } from '../common';
 import { HMSException } from '../error/HMSException';
 import { DeviceMap, SelectedDevices } from '../interfaces';
 import { HMSTrackSettings } from '../media/settings/HMSTrackSettings';
@@ -232,6 +233,16 @@ export default class AnalyticsEventFactory {
       name: 'subscriber.stats',
       level: AnalyticsEventLevel.INFO,
       properties,
+    });
+  }
+
+  static getKrispUsage() {
+    const duration = pluginUsageTracker.getPluginUsage('HMSKrispPlugin');
+    pluginUsageTracker.cleanup();
+    return new AnalyticsEvent({
+      name: 'krisp.usage',
+      level: AnalyticsEventLevel.INFO,
+      properties: { duration },
     });
   }
 
--- a/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
+++ b/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
@@ -235,6 +235,15 @@ export default class AnalyticsEventFactory {
     });
   }
 
+  static pluginStateChanged(name: string) {
+    return new AnalyticsEvent({
+      name: 'plugin.state.changed',
+      timestamp: new Date().getTime(),
+      properties: { pluginName: name },
+      level: AnalyticsEventLevel.INFO,
+    });
+  }
+
   private static eventNameFor(name: string, ok: boolean) {
     const suffix = ok ? 'success' : 'failed';
     return `${name}.${suffix}`;
--- a/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
+++ b/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
@@ -236,8 +236,8 @@ export default class AnalyticsEventFactory {
     });
   }
 
-  static getKrispUsage() {
-    const duration = pluginUsageTracker.getPluginUsage('HMSKrispPlugin');
+  static getKrispUsage(sessionID: string) {
+    const duration = pluginUsageTracker.getPluginUsage('HMSKrispPlugin', sessionID);
     return new AnalyticsEvent({
       name: 'krisp.usage',
       level: AnalyticsEventLevel.INFO,
--- a/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
+++ b/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
@@ -245,6 +245,14 @@ export default class AnalyticsEventFactory {
     });
   }
 
+  static transportLeave() {
+    return new AnalyticsEvent({
+      name: 'transport.leave',
+      level: AnalyticsEventLevel.INFO,
+      properties: {},
+    });
+  }
+
   private static eventNameFor(name: string, ok: boolean) {
     const suffix = ok ? 'success' : 'failed';
     return `${name}.${suffix}`;
--- a/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
+++ b/packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
@@ -235,15 +235,6 @@ export default class AnalyticsEventFactory {
     });
   }
 
-  static pluginStateChanged(name: string) {
-    return new AnalyticsEvent({
-      name: 'plugin.state.changed',
-      timestamp: new Date().getTime(),
-      properties: { pluginName: name },
-      level: AnalyticsEventLevel.INFO,
-    });
-  }
-
   private static eventNameFor(name: string, ok: boolean) {
     const suffix = ok ? 'success' : 'failed';
     return `${name}.${suffix}`;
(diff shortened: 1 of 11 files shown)


packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -100,6 +106,7 @@ export class HMSAudioPluginsManager {
       this.pluginsMap.set(name, plugin);
       await this.processPlugin(plugin);
       await this.connectToDestination();
+      this.pluginCallback?.(name, Date.now(), 'added');
```
KaustubhKumar05, line 109: Passing Date.now() is probably redundant, will do it inside the method only

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -235,6 +235,15 @@ export default class AnalyticsEventFactory {
     });
   }
 
+  static pluginStateChanged(name: string) {
+    return new AnalyticsEvent({
+      name: 'plugin.state.changed',
+      timestamp: new Date().getTime(),
```
raviteja83, line 241: why this extra timestamp? Date.now() would do no or we need in seconds?

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -40,11 +41,13 @@ export class HMSAudioPluginsManager {
   // This will replace the native track in peer connection when plugins are enabled
   private outputTrack?: MediaStreamTrack;
   private pluginAddInProgress = false;
+  private eventBus;
```
raviteja83, line 44: I think you can just make it private in the constructor. it is the same

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -138,6 +142,7 @@ export class HMSAudioPluginsManager {
   }
 
   async removePlugin(plugin: HMSAudioPlugin) {
+    this.eventBus.analytics.publish(AnalyticsEventFactory.pluginStateChanged(plugin.getName()));
```
raviteja83, line 145: removePluginInternal already does this I think

packages/hms-video-store/src/sdk/store/Store.ts
```diff
@@ -50,11 +51,40 @@ class Store {
   private userAgent: string = createUserAgent(this.env);
   private polls = new Map<string, HMSPoll>();
   private whiteboards = new Map<string, HMSWhiteboard>();
+  private pluginUsage: Record<string, number> = { HMSKrispPlugin: 0 };
```
raviteja83, line 54: use a map instead

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -0,0 +1,38 @@
...
+      this.pluginLastAddedAt.set(name, addedAt);
+      console.log('krisp last added at', addedAt);
+    } else if (event.name === 'mediaPlugin.stats') {
+      const name = event.properties.plugin_name;
+      const duration = event.properties.duration;
+      this.pluginUsage.set(name, (this.pluginUsage.get(name) || 0) + duration);
+    }
+  };
+
+  reset = () => {
```
raviteja83, line 32: should be cleanUp similar to other classes. where is this called?

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -0,0 +1,38 @@
...
+    return this.pluginUsage.get(name);
+  };
+
+  updatePluginUsage = (event: AnalyticsEvent) => {
+    if (event.name === 'mediaPlugin.added') {
+      const name = event.properties.plugin_name;
+      const addedAt = event.properties.added_at;
+      this.pluginLastAddedAt.set(name, addedAt);
+      console.log('krisp last added at', addedAt);
+    } else if (event.name === 'mediaPlugin.stats') {
```
raviteja83, line 25: I belive this is not trigerred if you directly leave. see if we don't miss this

KaustubhKumar05: This is handled in getPluginUsage. Storing the lastAddedAt for this reason

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  updatePluginUsage = (event: AnalyticsEvent) => {
+    if (event.name === 'mediaPlugin.added') {
+      const name = event.properties.plugin_name;
+      const addedAt = event.properties.added_at;
+      this.pluginLastAddedAt.set(name, addedAt);
+    } else if (event.name === 'mediaPlugin.stats') {
+      const name = event.properties.plugin_name;
+      const duration = event.properties.duration;
+      this.pluginUsage.set(name, (this.pluginUsage.get(name) || 0) + duration * 1000);
```
raviteja83, line 27: should we add the duration if usage.get() is undefined?

packages/hms-video-store/src/transport/index.ts
```diff
@@ -495,6 +497,14 @@ export default class HMSTransport {
     this.joinParameters = undefined;
     HMSLogger.d(TAG, 'leaving in transport');
     try {
+      this.eventBus.analytics.publish(
+        new AnalyticsEvent({
```
raviteja83, line 501: we can add this in AnalyticsEventFactory I think. for consistency

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -235,6 +236,16 @@ export default class AnalyticsEventFactory {
     });
   }
 
+  static getKrispUsage() {
+    const duration = pluginUsageTracker.getPluginUsage('HMSKrispPlugin');
+    pluginUsageTracker.cleanup();
```
raviteja83, line 241: cleanup should only be done for the key no? probably this is not the right place to call. we can call separately where. all other cleanup's are called on leave

packages/hms-video-store/src/plugins/audio/AudioPluginsAnalytics.ts
```diff
@@ -24,6 +24,7 @@ export class AudioPluginsAnalytics {
     this.addedTimestamps[name] = Date.now();
     this.initTime[name] = 0;
     this.pluginSampleRate[name] = sampleRate;
+    this.eventBus.analytics.publish(MediaPluginsAnalyticsFactory.added(name, this.addedTimestamps[name]));
```
raviteja83, line 27: did we check if this is getting sent to backend as well.

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+  };
+
+  cleanup = () => {
+    this.pluginUsage.clear();
+    this.pluginLastAddedAt.clear();
+  };
+}
+
+export const pluginUsageTracker = new PluginUsageTracker();
```
raviteja83, line 39: we should probably avoid singleton's. with proctoring we are going to use multiple rooms. if all of them have the same plugin this will fail.

packages/hms-video-store/src/plugins/audio/AudioPluginsAnalytics.ts
KaustubhKumar05: Yup
<br />

packages/hms-video-store/src/common/PluginUsageTracker.ts
KaustubhKumar05: Can make the key sessionID-pluginName instead of just pluginName. Since this object is used to send the value before leave, creating a new one will lead to data loss

KaustubhKumar05: And we only send the event once/session

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -0,0 +1,53 @@
...
+      const extraDuration = lastAddedAt ? Date.now() - lastAddedAt : 0;
+      this.pluginUsage.set(pluginKey, (this.pluginUsage.get(pluginKey) || 0) + extraDuration);
+      this.pluginLastAddedAt.delete(pluginKey);
+    }
+    return this.pluginUsage.get(pluginKey);
+  };
+
+  updatePluginUsageData = (event: AnalyticsEvent, sessionID: string) => {
+    // Sent on leave, after krisp usage is sent
+    if (event.name === 'subscriber.stats') {
```
raviteja83, line 24: do we not have an event for leave?

raviteja83: not sure if it's a good idea though we would send this always.

KaustubhKumar05: I can add a new event - "Leave initiated" and can listen to that. The new event should be fired before event bus and other classes are cleaned up. Does this work @eswarclynn ?

raviteja83 (approved):

PR #2687, opened 2024-03-11
fix: radio/checkbox shrink for longer options

KaustubhKumar05:
image

### Fixed:

- Spacing parity between checkbox - option and radio - option
- Radio/checkbox shrinkage when the option is large
- Max limit for option input


raviteja83:
@KaustubhKumar05 should be chckbox/radio be at the start or in the center?

PR #2694, opened 2024-03-12
fix: design parity for hand raise actions

KaustubhKumar05:


raviteja83 (approved):

packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
```diff
@@ -324,21 +323,16 @@ const HandRaisedAccordionParticipantActions = ({ peerId, role }: { peerId: strin
     peerId,
     role,
   });
+  const quickActionStyle = { p: '$1', borderRadius: '$round' };
```
raviteja83, line 326: we could move this out of the component. as it is not dependent on anything

raviteja83 (approved):

PR #2695, opened 2024-03-12
fix: pause play on click event

amar-1995:


packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -336,15 +336,23 @@ const HLSView = () => {
...
-  const onClickHandler = useCallback(() => {
+  const onClickHandler = useCallback(async () => {
+    if (!(isMobile || isLandscape) && hlsState?.variants[0]?.playlist_type !== HLSPlaylistType.DVR) {
+      return;
+    }
     if (!(isMobile || isLandscape)) {
+      if (isPaused) {
+        await hlsPlayer?.play();
+      } else {
+        hlsPlayer.pause();
```
raviteja83, line 347: no conditional(?.) here?

packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -336,15 +336,23 @@ const HLSView = () => {
     },
     [hlsState?.variants, isLandscape, isMobile, onSeekTo],
   );
-  const onClickHandler = useCallback(() => {
+  const onClickHandler = useCallback(async () => {
+    if (!(isMobile || isLandscape) && hlsState?.variants[0]?.playlist_type !== HLSPlaylistType.DVR) {
```
raviteja83, line 340: this conditions seems to be common. we can use tsPattern here but we can reuse this condition otherwise also

packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -337,21 +337,27 @@ const HLSView = () => {
...
-    }
+    match({ isMobile, isLandscape, playlist_type: hlsState?.variants[0]?.playlist_type })
+      .with({ playlist_type: HLSPlaylistType.DVR, isMobile: false, isLandscape: false }, async () => {
+        if (isPaused) {
+          await hlsPlayer?.play();
+        } else {
+          hlsPlayer?.pause();
+        }
+      })
+      .with({ isMobile: true }, () => {
```
raviteja83, line 348: you can add  { isLandscape: true } as argument. as the values are same.

PR #2700, opened 2024-03-13
fix: added play pause state for desktop

amar-1995:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/PlayPauseSeekControls.tsx
```diff
@@ -0,0 +1,158 @@
...
+  hoverControlsVisible: {
+    seekBackward: boolean;
+    seekForward: boolean;
+    pausePlay: boolean;
+  };
+}) => {
+  const isMobile = useMedia(config.media.md);
+  const isLandscape = useIsLandscape();
+
+  if (!isMobile && !isLandscape) {
```
raviteja83, line 61: why is this check here? if we are anyways having this check, we can combine above and this component

amar-1995: Both components are different.
One is used to show icon on player controls. it will always be visible. Second one is hovering over the video view.

raviteja83: both have almost similar code. can we use variants to make it simpler.

packages/roomkit-react/src/Prebuilt/components/HMSVideo/PlayPauseSeekControls.tsx
```diff
@@ -0,0 +1,158 @@
...
+export const PlayPauseSeekControls = ({
+  isPaused,
+  onSeekTo,
+}: {
+  isPaused: boolean;
+  onSeekTo: (value: number) => void;
+}) => {
+  return (
+    <>
+      <SeekControls
```
raviteja83, line 21: This component should be SeekControl not Controls as it a single component

packages/roomkit-react/src/Prebuilt/components/HMSVideo/PlayPauseSeekControls.tsx
```diff
@@ -0,0 +1,158 @@
...
+        bg: '#00000066',
+        display: 'inline-flex',
+        gap: '$2',
+        zIndex: 1,
+        size: '100%',
+        visibility: controlsVisible ? `` : `hidden`,
+        opacity: controlsVisible ? `1` : '0',
+      }}
+    >
+      <SeekControls
```
raviteja83, line 130: The below section here and the above component render looks almost same.

packages/roomkit-react/src/Prebuilt/components/HMSVideo/PlayPauseSeekControls.tsx
```diff
@@ -0,0 +1,158 @@
...
+};
+
+// overlay handlers
+export const PlayPauseSeekOverlayControls = ({
+  isPaused,
+  controlsVisible,
+  hoverControlsVisible,
+}: {
+  isPaused: boolean;
+  controlsVisible: boolean;
```
raviteja83, line 51: showControls

raviteja83 (approved):

PR #2712, opened 2024-03-17
fix: chat count in landscape and fullscreen

amar-1995:
chat count
<br /> chat with count <br />
old image
<br /> Old without chat icon <br />
new image
<br /> new with chat icon in fullscreen  <br />

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
+++ b/packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
@@ -592,7 +592,7 @@ const HLSView = () => {
                       }}
                     >
                       <HMSVideoPlayer.Controls.Right>
-                        {isLandscape && showChat && (
+                        {(isLandscape || (isMobile && isFullScreen)) && showChat && (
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -589,7 +592,16 @@ const HLSView = () => {
                       }}
                     >
                       <HMSVideoPlayer.Controls.Right>
-                        {isLandscape && <ChatToggle />}
+                        {(isLandscape || (isMobile && isFullScreen)) && showChat && (
+                          <ChatToggle
+                            onClick={() => {
+                              if (isFullScreen) {
+                                toggle();
```
raviteja83, line 599: why are we toggling fullscreen? because we cannot add sidepane here?

raviteja83: one way is to make the conferencing container fullscreen.

raviteja83 (approved):

PR #2713, opened 2024-03-18
fix: update pluginUsageTracker implementation

KaustubhKumar05:
- [x] Combine usage across device change, plugin addition/removal
- [x] Should be session/room specific to accommodate multiple rooms
- [x] Toggle methods inside HMSKrispPlugin
- [x] Send the value via `krisp.usage` event before leave


packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -19,25 +19,33 @@ class PluginUsageTracker {
...
+      case 'mediaPlugin.stats': {
+        if (this.pluginLastAddedAt.has(pluginKey)) {
+          const duration =
+            event.properties.duration || (Date.now() - (this.pluginLastAddedAt.get(pluginKey) || 0)) / 1000;
+          this.pluginUsage.set(pluginKey, (this.pluginUsage.get(pluginKey) || 0) + Math.max(duration, 0) * 1000);
+          this.pluginLastAddedAt.delete(pluginKey);
+        }
       }
     }
+    return;
```
raviteja83, line 48: doesn't seem necessary.

packages/hms-video-store/src/common/PluginUsageTracker.ts
```diff
@@ -19,25 +19,33 @@ class PluginUsageTracker {
...
+        return;
+      }
+      case 'mediaPlugin.toggled.on':
+      case 'mediaPlugin.added': {
+        const addedAt = event.properties.added_at || Date.now();
+        this.pluginLastAddedAt.set(pluginKey, addedAt);
+        break;
+      }
+      case 'mediaPlugin.toggled.off':
+      case 'mediaPlugin.stats': {
```
raviteja83, line 39: add a default case as well. do nothing

packages/hms-video-store/src/plugins/video/HMSVideoPlugin.ts
```diff
@@ -29,7 +29,7 @@ export interface HMSVideoPlugin {
    * variables, loading ML models etc. This can be used by a plugin to ensure it's prepared at the time
    * processVideoFrame is called.
    */
-  init(): Promise<void>;
+  init(sessionID?: string): Promise<void>;
```
raviteja83, line 32: we should call init with sessionID also no

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -87,6 +87,9 @@ export class HMSAudioPluginsManager {
     }
 
     await this.validateAndThrow(name, plugin);
+    // @ts-ignore
+    plugin.setEventBus?.(this.eventBus);
+    // @ts-ignore
```
raviteja83, line 92: not needed

packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
```diff
@@ -145,6 +145,10 @@ const NoiseCancellation = () => {
           onClick={e => e.stopPropagation()}
           onCheckedChange={value => {
             setActive(value);
+            if (isPluginAdded) {
+              plugin.toggle();
+              console.log(plugin, 'lggr');
```
raviteja83, line 150: not needed

packages/hms-video-store/src/transport/index.ts
```diff
@@ -495,9 +498,10 @@ export default class HMSTransport {
     this.joinParameters = undefined;
     HMSLogger.d(TAG, 'leaving in transport');
     try {
-      const sessionID = this.store.getRoom()?.sessionId || '';
-      this.eventBus.analytics.publish(AnalyticsEventFactory.getKrispUsage(sessionID));
-      this.eventBus.analytics.publish(AnalyticsEventFactory.transportLeave());
+      this.eventBus.analytics.publish(
+        AnalyticsEventFactory.getKrispUsage(this.pluginUsageTracker.getPluginUsage('HMSKrispPlugin')!),
+      );
+      this.pluginUsageTracker.updatePluginUsageData(transportLeaveEvent);
```
raviteja83, line 504: is this required if we are doing the above?

KaustubhKumar05: Currently yes, this is the cleanup. Can combine the calculation + cleanup, was initially kept separate to accommodate the singleton approach. Will remove the extra event too

raviteja83 (approved):

PR #2714, opened 2024-03-18
fix: rerender chat when opened with existing messages

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Chat/ChatBody.tsx
```diff
@@ -410,6 +410,7 @@ export const ChatBody = React.forwardRef<VariableSizeList, { scrollToBottom: (co
     }, [blacklistedMessageIDs, messages]);
 
     const vanillaStore = useHMSVanillaStore();
+    const rerenderOnFirstMount = useRef(false);
```
raviteja83, line 413: why not do this in an effect with no dependency? that will also run only once no?

KaustubhKumar05: Need to wait listRef and filteredMessages to be defined. Had initially tried with no deps, but it executes before both are defined

raviteja83 (approved):

PR #2715, opened 2024-03-18
feat: hand raise config

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/hooks/useMetadata.tsx
```diff
@@ -15,7 +15,7 @@ export const useMyMetadata = () => {
   const metaData = useHMSStore(selectPeerMetadata(localPeerId));
   const isHandRaised = useHMSStore(selectHasPeerHandRaised(localPeerId));
 
-  const update = async updatedFields => {
+  const update = async (updatedFields: Record<string, any>) => {
```
raviteja83, line 18: I think we should if handraise has to be supported here. check with product once.

raviteja83, line 18: anyway we can get rid of this hook? how many places is it used?

raviteja83 (approved):

PR #2725, opened 2024-03-20
feat: screenshare optimization

amar-1995:
It will optimize screenshare to use the pixels as initialized in publish param


packages/hms-video-store/src/sdk/LocalTrackManager.ts
```diff
@@ -187,6 +222,8 @@ export class LocalTrackManager {
       HMSLogger.d('retrieving screenshare with ', { config }, { constraints });
       // @ts-ignore [https://github.com/microsoft/TypeScript/issues/33232]
       stream = (await navigator.mediaDevices.getDisplayMedia(constraints)) as MediaStream;
+      // todo change stream logic
+      await this.optimizeScreenShareConstraint(stream, constraints);
```
raviteja83, line 226: should this be opt-in instead of for overyone? what is the impact on current behaviour? is the output consistent?

amar-1995: no, will updating it. will get init value to ally it or not.

packages/hms-video-store/src/sdk/LocalTrackManager.ts
```diff
@@ -158,8 +158,38 @@ export class LocalTrackManager {
...
+          videoConstraint.height = videoHeight / sqrt_ratio;
+        } else {
+          videoConstraint.height = videoHeight * sqrt_ratio;
+          videoConstraint.width = videoWidth * sqrt_ratio;
+        }
+        await stream.getVideoTracks()[0].applyConstraints(videoConstraint);
+      }
+    });
+  }
+  async getLocalScreen(partialConfig?: HMSScreenShareConfig, needOptimized = false) {
```
raviteja83, line 192: optimise = false

KaustubhKumar05 (approved):

PR #2748, opened 2024-03-26
feat: add type to peer

raviteja83:

-
-


packages/hms-video-store/src/sdk/models/peer/peer.test.ts
```diff
@@ -45,6 +46,7 @@ describe('HMSLocalPeer', () => {
     name: 'John Doe',
     role: getParamsForRole(role),
     customerUserId: userId,
+    type: 'regular' as HMSPeerType,
```
eswarclynn, line 49: Do we need the 'as' here? Because I don't see it below at line 85.

eswarclynn (approved):

PR #2761, opened 2024-03-26
feat: hand raise config

raviteja83:

-
-


packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/MwebOptions.tsx
```diff
@@ -175,7 +171,7 @@ export const MwebOptions = ({
               </ActionTile.Root>
             )}
 
-            {!noAVPermissions ? (
+            {elements.hand_raise ? (
```
amar-1995, line 174: av permission is not needed here?

raviteja83: no, earlier we did not have any configuration hence that was used as a placeholder

amar-1995 (approved):

PR #2763, opened 2024-03-27
fix: scroll to latest chat message

amar-1995:

-
-

### DIFF
--- a/packages/hms-video-store/src/reactive-store/HMSSDKActions.ts
+++ b/packages/hms-video-store/src/reactive-store/HMSSDKActions.ts
@@ -415,7 +415,8 @@ export class HMSSDKActions<T extends HMSGenericTypes = { sessionStore: Record<st
       HMSLogger.w('sendMessage', 'Failed to send message', messageInput);
       throw Error(`sendMessage Failed - ${JSON.stringify(messageInput)}`);
     }
-    if (!!messageInput.type && this.ignoredMessageTypes.includes(messageInput.type)) {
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/reactive-store/HMSSDKActions.ts
```diff
@@ -429,7 +429,11 @@ export class HMSSDKActions<T extends HMSGenericTypes = { sessionStore: Record<st
       senderRole: localPeer?.role?.name,
       ignored: !!messageInput.type && this.ignoredMessageTypes.includes(messageInput.type),
     };
-    this.putMessageInStore(hmsMessage);
+    // update directly to store without batch
+    this.setState(store => {
```
raviteja83, line 433: we should not do this for ignored messages. should do a return if ignored.

amar-1995: yes,

packages/hms-video-store/src/reactive-store/HMSSDKActions.ts
```diff
@@ -415,6 +415,9 @@ export class HMSSDKActions<T extends HMSGenericTypes = { sessionStore: Record<st
       HMSLogger.w('sendMessage', 'Failed to send message', messageInput);
       throw Error(`sendMessage Failed - ${JSON.stringify(messageInput)}`);
     }
+    if (!!messageInput.type && this.ignoredMessageTypes.includes(messageInput.type)) {
```
raviteja83, line 418: can we store this in a var and reuse below

raviteja83 (approved):

PR #2765, opened 2024-03-27
fix: cumulative values for packets lost and sent

eswarclynn:

-
-


raviteja83 (approved):

packages/hms-video-store/src/analytics/stats/PublishStatsAnalytics.ts
```diff
@@ -133,8 +133,8 @@ class RunningLocalTrackAnalytics extends RunningTrackAnalytics {
       avg_available_outgoing_bitrate_bps: this.calculateAverage('availableOutgoingBitrate'),
       avg_bitrate_bps: this.calculateAverage('bitrate'),
       avg_fps: this.calculateAverage('framesPerSecond'),
-      total_packets_lost: this.calculateDifferenceForSample('packetsLost'),
-      total_packets_sent: this.calculateDifferenceForSample('packetsSent'),
+      total_packets_lost: this.getLatestStat().packetsLost,
```
raviteja83, line 136: this will not be null/undefined right?

eswarclynn: In reality it won't but the type is nullable only and it's handled

PR #2772, opened 2024-04-01
fix: auto-open vb sidepane if default is present only in preview

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/AppData/AppData.tsx
```diff
@@ -112,19 +123,12 @@ export const AppData = React.memo(() => {
...
-        defaultMediaURL = media.url;
-      }
-    });
-    if (defaultMediaURL) {
+    if (defaultMediaURL && !sidepaneOpenedRef.current && !isMobile && isLocalVideoEnabled) {
       hmsActions.setAppData(APP_DATA.background, defaultMediaURL);
-      if (isLocalVideoEnabled && !isMobile) {
-        toggleVB();
-      }
+      sidepaneOpenedRef.current = true;
```
raviteja83, line 128: we can check this from appData itself no?

raviteja83 (approved):

PR #2782, opened 2024-04-04
fix: added min width and height

amar-1995:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/HMSVideo.jsx
```diff
@@ -44,6 +44,8 @@ export const HMSVideo = forwardRef(({ children, ...props }, videoRef) => {
           height: 'auto',
           maxWidth: '100%',
           maxHeight: '100%',
+          minWidth: '100%',
```
raviteja83, line 47: one dimension should be fixed and the other should be auto. this is same as height:  '100%, width: '100%' which we should not be doing

raviteja83 (approved):

PR #2785, opened 2024-04-04
fix: reset whiteboard token on role change

eswarclynn:

-
-


packages/hms-video-store/src/session-store/interactivity-center/HMSWhiteboardCenter.ts
```diff
@@ -72,6 +72,29 @@ export class WhiteboardInteractivityCenter implements HMSWhiteboardInteractivity
     this.listener = listener;
   }
 
+  async handleLocalRoleUpdate() {
+    const whiteboards = this.store.getWhiteboards();
+
+    for await (const whiteboard of whiteboards.values()) {
```
raviteja83, line 78: looks odd. why is wait needed here?

eswarclynn: We're using await inside the for loop implementation

raviteja83 (approved):

PR #2799, opened 2024-04-08
fix: transcription text color

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/HMSVideo.jsx
```diff
@@ -17,7 +17,8 @@ export const HMSVideo = forwardRef(({ children, ...props }, videoRef) => {
           },
         },
         '& video::cue': {
-          color: 'white',
+          // default for on-surface-high
+          color: '#EFF0FA',
```
raviteja83, line 21: $on_surface_hight doesn't work?

KaustubhKumar05: The color should be independent of the theme

raviteja83 (approved):

PR #2807, opened 2024-04-10
feat: hls multiple variants

amar-1995:


packages/hms-video-store/src/notification-manager/managers/RoomUpdateManager.ts
```diff
@@ -149,19 +149,32 @@ export class RoomUpdateManager {
...
+      }
     });
     return hls;
   }
   private updateHLSStatus(notification: HLSNotification) {
     const room = this.store.getRoom();
     const running =
       notification.variants && notification.variants.length > 0
-        ? this.isStreamingRunning(notification.variants[0].state)
+        ? !notification.variants.find(variant => this.isStreamingRunning(variant.state))
```
raviteja83, line 177: use `.some` instead of find

packages/hms-video-store/src/interfaces/room.ts
```diff
@@ -110,4 +110,5 @@ export interface HLSVariant {
   startedAt?: Date;
   initialisedAt?: Date;
   state?: HMSStreamingState;
+  stream_type?: string;
```
raviteja83, line 113: this is not any string right? `regular | screen | composite` i think

amar-1995: yes

raviteja83 (approved):

PR #2827, opened 2024-04-16
fix: call effects plugin method once the plugin is loaded

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBOption.tsx
```diff
@@ -24,11 +26,17 @@ const Root = ({
       borderRadius: '$1',
       bg: '$surface_bright',
       border: `4px solid ${isActive ? '$primary_default' : '$surface_dim'}`,
+      opacity: disabled ? 0.8 : 1,
```
raviteja83, line 29: we should cursor: not-allowed

packages/hms-virtual-background/src/HMSEffectsPlugin.ts
```diff
@@ -28,33 +32,62 @@ export class HMSEffectsPlugin implements HMSMediaStreamPlugin {
...
     this.backgroundType = HMSVirtualBackgroundTypes.BLUR;
-    this.effects.setBlur(blur);
+    this.removeBackground();
+    await this.executeAfterInit(() => {
+      this.effects.setBlur(this.blurAmount);
+    });
   }
 
   async setPreset(preset: string) {
     this.preset = preset;
```
raviteja83, line 87: we should check if it is available. They don't have a type for presets? better to use specific type here

packages/hms-virtual-background/src/HMSEffectsPlugin.ts
```diff
@@ -28,33 +32,62 @@ export class HMSEffectsPlugin implements HMSMediaStreamPlugin {
...
-    this.effects.clearBackground();
+    this.executeAfterInit(() => {
+      this.effects.clearBackground();
+    });
   }
 
-  setBlur(blur: number) {
-    this.removeBackground();
+  async setBlur(blur: number) {
     this.blurAmount = blur;
```
raviteja83, line 78: we should also mention a range or add a check here.

packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBPicker.tsx
```diff
@@ -89,19 +94,30 @@ export const VBPicker = ({ backgroundMedia = [] }: { backgroundMedia: VirtualBac
...
+    return () => setLoadingEffects(false);
+  }, [isVideoOn, setLoadingEffects, toggleVB]);
 
   return (
     <Flex css={{ pr: '$6', size: '100%' }} direction="column">
       <Flex align="center" justify="between" css={{ w: '100%', background: '$surface_dim', pb: '$4' }}>
-        <Text variant="h6" css={{ color: '$on_surface_high' }}>
-          Virtual Background
+        <Text variant="h6" css={{ color: '$on_surface_high', display: 'flex', alignItems: 'center' }}>
+          Virtual Background {isMobile && loadingEffects ? <Loading size={18} style={{ marginLeft: '0.5rem' }} /> : ''}
```
raviteja83, line 120: css would work with Loader no?

packages/hms-virtual-background/src/HMSEffectsPlugin.ts
KaustubhKumar05: image
Nope, they just use `string`. The doc is outdated, error message says `balanced` and `quality` are the only permitted values - we were using `lightning` before this

packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBPicker.tsx
KaustubhKumar05: Nope, it is a standard SVG:

```jsx
type LoadingProps = Props & React.SVGProps<SVGSVGElement>;
```

packages/hms-virtual-background/src/HMSEffectsPlugin.ts
raviteja83: let's add that to the type. 'balanced' | 'quality' instead of string

raviteja83 (approved):

PR #2847, opened 2024-04-22
fix: handle emoji reactions for large rooms

KaustubhKumar05:


packages/hms-video-store/src/reactive-store/HMSSDKActions.ts
```diff
@@ -602,6 +603,12 @@ export class HMSSDKActions<T extends HMSGenericTypes = { sessionStore: Record<st
     await this.sdk.lowerRemotePeerHand(peerId);
   }
 
+  async getPeer(peerId: getPeerRequestParams) {
+    const peer = await this.sdk.getPeer(peerId);
+    const hmsPeer = SDKToHMS.convertPeer(peer) as HMSPeer;
```
raviteja83, line 608: directly return the convertPeer

packages/hms-video-store/src/sdk/index.ts
```diff
@@ -728,6 +729,11 @@ export class HMSSdk implements HMSInterface {
     return await this.sendMessageInternal({ message, recipientPeer, type });
   }
 
+  async getPeer(peerID: getPeerRequestParams) {
+    const response = await this.transport.signal.getPeer(peerID);
+    return createRemotePeer(response, this.store);
```
raviteja83, line 734: we should add a check here if response is undefined or not

raviteja83 (approved):

PR #2851, opened 2024-04-25
fix: mweb speaker selection

raviteja83:

-
-

### DIFF
--- a/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
+++ b/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
@@ -198,7 +198,7 @@ export class AudioSinkManager {
   private handleAudioDeviceChange = (event: HMSDeviceChangeEvent) => {
     // this means the initial load
     if (!event.selection) {
-      console.log('device change called');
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -286,8 +290,7 @@ export class AudioSinkManager {
    */
   // eslint-disable-next-line complexity
   private autoSelectAudioOutput = async () => {
-    if (!this.audioSink?.children.length) {
-      HMSLogger.d(this.TAG, 'No remote audio added yet');
+    if ('ondevicechange' in navigator.mediaDevices) {
```
KaustubhKumar05, line 293: This is early return to honor manual selection?

KaustubhKumar05 (approved):

packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
raviteja83: to not trigger in desktop. this is only needed when device change not supportes

PR #2859, opened 2024-04-29
fix: remove pdf state after session close

amar-1995:


packages/roomkit-react/src/Prebuilt/layouts/PDFView.jsx
```diff
@@ -11,18 +11,28 @@ import { usePDFConfig, useResetPDFConfig } from '../components/AppData/useUISett
...
-
+  const [isPDFShared, setIsPDFShared] = useState(true);
   // need to send resetConfig to clear configuration, if stop screenshare occurs.
   const { iframeRef, startPDFShare, isPDFShareInProgress } = usePDFShare(resetConfig);
 
+  useEffect(() => {
+    // no working in other useEffect, as return is called multiple time on state change
+    return () => {
+      if (isPDFShared) {
+        resetConfig();
```
raviteja83, line 22: do we need to maintain a state? if it not shared, calling reset would anyways do nothing right? if this component is open that means some state is already on no?

amar-1995: There are three state:
1. Here we initialize the pdf config, with a pdf file. We check that config is present in `VideoStreamingSection`
2. this state we get from `usePDFShare` which is `isPDFShareInProgress`, means pdf is not started and it will be in popup mode.
3. Here we maintain state of PDF sharing started. This is to maintain and clear after the if session closes and restarted.

raviteja83 (approved):

PR #2868, opened 2024-05-02
fix: remove unused components

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/HMSVideo/PlayButton.tsx
raviteja83, line 1: this is not used anywhere?

KaustubhKumar05: image

Nope, I think this was added for HLS UI update but not used
cc @amar-1995

raviteja83 (approved):

PR #2872, opened 2024-05-06
fix: active state for poll, chat, peer list CTAs

KaustubhKumar05:
- [x] Remove autofocus for participant search
 - [x] Active state for chat and participant list CTA


packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
```diff
@@ -471,6 +472,8 @@ export const ParticipantSearch = ({
         }}
         autoComplete="off"
         aria-autocomplete="none"
+        // @ts-ignore Flag added for customiser UI
+        autoFocus={!elements?.chat?.disable_autofocus}
```
raviteja83, line 476: this is added where? from the customiser side? can we pass in participants only?

raviteja83 (approved):

PR #2884, opened 2024-05-09
feat: custom ice servers in join config

KaustubhKumar05:
- [x] Update join config
- [x] Use ICE servers 
- [ ] Testing


packages/hms-video-store/src/signal/init/index.ts
```diff
@@ -26,26 +28,26 @@ export default class InitService {
     userAgent,
     initEndpoint = 'https://prod-init.100ms.live',
     region = '',
+    iceServers = [],
```
raviteja83, line 31: why are we adding the default here?

packages/hms-video-store/src/signal/init/index.ts
```diff
@@ -78,9 +80,12 @@ export function getUrl(endpoint: string, peerId: string, userAgent: string, regi
...
 }
 
-export function transformInitConfig(config: any): InitConfig {
+export function transformInitConfig(config: any, iceServers?: HMSICEServer[]): InitConfig {
   return {
     ...config,
-    rtcConfiguration: { ...config.rtcConfiguration, iceServers: config.rtcConfiguration?.ice_servers },
+    rtcConfiguration: {
+      ...config.rtcConfiguration,
+      iceServers: transformIceServerConfig(iceServers) || config.rtcConfiguration?.ice_servers,
```
raviteja83, line 88: pass the config also to the function and get the config ice_servers if joinConfig ones are empty

packages/hms-video-store/src/signal/init/index.ts
raviteja83: add in the last place in the chain

raviteja83 (approved):

PR #2888, opened 2024-05-09
feat: sdk - start/stop events for transcriptions

amar-1995:

-
-


packages/hms-video-store/src/notification-manager/managers/RoomUpdateManager.ts
```diff
@@ -125,12 +128,18 @@ export class RoomUpdateManager {
     if (!transcriptions) {
       return [];
     }
-    return transcriptions.map((transcription: TranscriptionNotification) => {
+    const output = transcriptions.map((transcription: TranscriptionNotification) => {
```
raviteja83, line 131: this extra var does not make any difference

packages/hms-video-store/src/reactive-store/adapter.ts
```diff
@@ -139,10 +139,11 @@ export class SDKToHMS {
   }
 
   static convertRoom(sdkRoom: sdkTypes.HMSRoom, sdkLocalPeerId?: string): Partial<HMSRoom> {
-    const { recording, rtmp, hls } = SDKToHMS.convertRecordingStreamingState(
+    const { recording, rtmp, hls, transcriptions } = SDKToHMS.convertRecordingStreamingState(
       sdkRoom?.recording,
```
raviteja83, line 143: if the type is non-null why are we adding `?.`. do check if it is essential. it might be there earlier

packages/hms-video-store/src/sdk/index.ts
```diff
@@ -1012,6 +1013,38 @@ export class HMSSdk implements HMSInterface {
     await this.transport?.signal.stopHLSStreaming();
   }
 
+  async startTranscription(params?: TranscriptionConfig) {
+    if (!this.localPeer) {
+      throw ErrorFactory.GenericErrors.NotConnected(
+        HMSAction.VALIDATION,
+        'No local peer present, cannot start HLS streaming',
```
raviteja83, line 1020: transcriptions

packages/hms-video-store/src/sdk/index.ts
```diff
@@ -1012,6 +1013,38 @@ export class HMSSdk implements HMSInterface {
     await this.transport?.signal.stopHLSStreaming();
   }
 
+  async startTranscription(params?: TranscriptionConfig) {
```
raviteja83, line 1016: why is it optional here? in the interface it is not

packages/hms-video-store/src/sdk/index.ts
```diff
@@ -1012,6 +1013,38 @@ export class HMSSdk implements HMSInterface {
...
+  }
+
+  async stopTranscription(params?: TranscriptionConfig) {
+    if (!this.localPeer) {
+      throw ErrorFactory.GenericErrors.NotConnected(
+        HMSAction.VALIDATION,
+        'No local peer present, cannot stop HLS streaming',
+      );
+    }
+    if (!params) {
```
raviteja83, line 1039: the check directly be params?.mode if params is supposed to be optional

packages/hms-video-store/src/reactive-store/adapter.ts
amar-1995: it can be null type. It is possible we won't get transcriptions object itself

raviteja83: I am talking about sdkRoom. It cannot be null right once you are in session? can the function be called with a null sdkRoom?

amar-1995: okay got it.

raviteja83 (approved):

raviteja83:
add a selector to read transcription state from room

PR #2897, opened 2024-05-14
fix: extra get poll responses calls

KaustubhKumar05:
- [x] Fix extra get poll responses calls
- [x] Update the state when local peer opens the voting view
- [x] Store responses in app data
- [x] Fix store not being updated 
- Regression testing
   - [x] Fix switching between polls/quizzes


packages/roomkit-react/src/Prebuilt/components/Polls/Voting/Voting.tsx
```diff
@@ -27,6 +27,15 @@ export const Voting = ({ id, toggleVoting }: { id: string; toggleVoting: () => v
...
   const showSingleView = poll?.type === 'quiz' && poll.state === 'started';
 
+  useEffect(() => {
+    const getResponses = async () => {
+      if (poll && actions.interactivityCenter) {
+        await actions.interactivityCenter.getPollResponses(poll, true);
+      }
+    };
+    getResponses();
+  }, [poll, actions.interactivityCenter]);
```
KaustubhKumar05, line 37: Add a ref here to prevent multiple calls

packages/roomkit-react/src/Prebuilt/components/Polls/Voting/QuestionCard.jsx
```diff
@@ -73,6 +78,7 @@ export const QuestionCard = ({
         duration: Date.now() - startTime.current,
       },
     ]);
+    setLocalPeerChoice({ skipped: false, option: index });
```
KaustubhKumar05, line 81: Update the app data state here, pass the response as an object.
In Voting.tsx, create a method to update responseMap that accepts a new response

packages/roomkit-react/src/Prebuilt/common/utils.js
```diff
@@ -142,22 +142,22 @@ export const getPeerResponses = (questions, peerid, userid) => {
...
-  for (let i = 0; i < peerResponses.length; i++) {
-    // If another peer has attempted, undefined changes to an empty array
-    if (peerResponses[i] === undefined || peerResponses[i].length === 0) {
-      // Backend question index starts at 1
-      return i + 1;
+export const getIndexToShow = responses => {
+  let lastAttemptedIndex = 0;
+
+  Object.keys(responses).forEach(key => {
+    const keyNum = parseInt(key);
```
raviteja83, line 154: value doesn't have key as a number?

KaustubhKumar05: image
This is a js file, it's inferring key as a string. I need it as a num for comparison in L155
```
if (keyNum > lastAttemptedIndex && responses[key]) {
```

KaustubhKumar05: You mean something like this?

```js
if (responses[key] && responses[key].questionIndex > lastAttemptedIndex) {
```

raviteja83 (approved):

PR #2911, opened 2024-05-20
feat: resizable input

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Chat/ChatFooter.tsx
```diff
@@ -102,6 +102,13 @@ export const ChatFooter = ({ onSend, children }: { onSend: (count: number) => vo
       }
     }
   }, [defaultSelection, selectedPeer, selectedRole, setRoleSelector, isMobile, isLandscapeHLSStream, elements?.chat]);
+
+  const resetInputHeight = () => {
+    if (inputRef.current) {
+      inputRef.current.style.height = `${Math.max(32, inputRef.current.value ? inputRef.current.scrollHeight : 0)}px`;
```
raviteja83, line 108: this also should be min of scrollHeight and 24*4 right?

raviteja83: but given the place we are calling this, can't we just set it to default input height?

KaustubhKumar05: 32 is the default height. Since I am reusing this for onBlur - someone might have typed and didn't send - I am considering the input value as well
image

packages/roomkit-react/src/Prebuilt/components/Chat/ChatFooter.tsx
```diff
@@ -102,6 +102,19 @@ export const ChatFooter = ({ onSend, children }: { onSend: (count: number) => vo
       }
     }
   }, [defaultSelection, selectedPeer, selectedRole, setRoleSelector, isMobile, isLandscapeHLSStream, elements?.chat]);
+
+  const resetInputHeight = () => {
```
raviteja83, line 106: we can create these fn's outside or use useCallback.

raviteja83 (approved):

PR #2914, opened 2024-05-20
fix: handle video getting paused on incoming call interruption

raviteja83:

-
-


packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -281,6 +281,7 @@ export class AudioSinkManager {
       (async () => {
         await this.deviceManager.init(true, false);
         await this.autoSelectAudioOutput();
+        this.unpauseAudioTracks();
```
KaustubhKumar05, line 284: Considering this runs every 5 seconds, will audio and video be out of sync or will the latest audio be playing?

eswarclynn (approved):

PR #2916, opened 2024-05-21
fix: show elements when chat is paused or peer is blocked

KaustubhKumar05:
- After:
image

image


packages/roomkit-react/src/Prebuilt/components/Chat/Chat.tsx
```diff
@@ -57,20 +61,27 @@ export const Chat = () => {
     >
       {isMobile && elements?.chat?.is_overlay && !streaming ? null : <PinnedMessage />}
       <ChatBody ref={listRef} scrollToBottom={scrollToBottom} />
-
-      <ChatPaused />
-      <ChatBlocked />
+      <Flex align="center" css={{ w: '100%', gap: '$2' }}>
+        <ChatPaused />
+        <ChatBlocked />
+        {streaming && (!isChatEnabled || isLocalPeerBlacklisted) && (
```
raviteja83, line 67: these are to be shown when blacklisted?

raviteja83 (approved):

PR #2923, opened 2024-05-22
fix: correct message for first peer

amar-1995:

-
-


KaustubhKumar05 (approved):

KaustubhKumar05:
If multiple non hosts join, they will still see the "You're the first to join" message @amar-1995?

PR #2925, opened 2024-05-23
fix: alignment of textarea

KaustubhKumar05:
The placeholder and icon were not aligned

image


packages/roomkit-react/src/Prebuilt/components/Chat/ChatFooter.tsx
```diff
@@ -210,20 +210,19 @@ export const ChatFooter = ({ onSend, children }: { onSend: (count: number) => vo
...
               minHeight: '$16',
               position: 'relative',
-              py: '$6',
+              py: isOverlayChat && isMobile ? '$2' : '$6',
               pl: '$8',
               flexGrow: 1,
               r: '$1',
               '@md': {
-                minHeight: 'unset',
-                h: '$14',
```
KaustubhKumar05, line 226: Removed the fix height since this can grow now

PR #2945, opened 2024-05-27
fix: bump effects version, add resolution change callback

KaustubhKumar05:
https://100ms.atlassian.net/browse/WEB-2854
https://100ms.atlassian.net/browse/WEB-2855

- [x] Testing
- [x] Bump version
- [x] Update canvas on resolution change
- [x] Add callback for resolution change
- [x] Remove webgpu dependency which breaks for Safari iOS


packages/hms-virtual-background/src/HMSEffectsPlugin.ts
```diff
@@ -98,6 +98,10 @@ export class HMSEffectsPlugin implements HMSMediaStreamPlugin {
     });
   }
 
+  setOnResolutionChange(callback: (width: number, height: number) => void) {
```
raviteja83, line 101: onResolutionChange

raviteja83 (approved):

PR #2946, opened 2024-05-28
fix: check feature flags before adding plugins

KaustubhKumar05:


packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -69,6 +72,17 @@ export class HMSAudioPluginsManager {
       throw err;
     }
 
+    switch (plugin.getName()) {
+      case 'HMSKrispPlugin':
+        if (!this.room?.isNoiseCancellationEnabled) {
+          const errorMessage = 'Krisp Noise Cancellation is not enabled for this room';
+          console.error(errorMessage);
```
raviteja83, line 79: remove test log

raviteja83: where does this error showup? will other plugins continue to be added?

packages/hms-video-store/src/plugins/video/HMSMediaStreamPluginsManager.ts
```diff
@@ -2,19 +2,34 @@ import { HMSMediaStreamPlugin } from './HMSMediaStreamPlugin';
...
   }
 
   addPlugins(plugins: HMSMediaStreamPlugin[]): void {
-    plugins.forEach(plugin => this.plugins.add(plugin));
+    plugins.forEach(plugin => {
+      switch (plugin.getName()) {
+        case 'HMSEffectsPlugin':
+          if (!this.room?.isEffectsEnabled) {
+            const errorMessage = 'Effects Virtual Background is not enabled for this room';
+            console.error(errorMessage);
```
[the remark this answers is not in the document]
raviteja83: we should probably throw an error if only this plugin is there. otherwise other plugins won't be added as well. similar in audio as well.

packages/hms-video-store/src/plugins/video/HMSMediaStreamPluginsManager.ts
```diff
@@ -2,19 +2,38 @@ import { HMSMediaStreamPlugin } from './HMSMediaStreamPlugin';
...
   }
 
   addPlugins(plugins: HMSMediaStreamPlugin[]): void {
-    plugins.forEach(plugin => this.plugins.add(plugin));
+    plugins.forEach(plugin => {
+      switch (plugin.getName()) {
+        case 'HMSEffectsPlugin':
+          if (!this.room?.isEffectsEnabled) {
+            const errorMessage = 'Effects Virtual Background is not enabled for this room';
+            if (this.getPlugins().length === 1) {
```
raviteja83, line 26: should we not check both this.plugins and plugins here?

packages/hms-video-store/src/plugins/video/HMSMediaStreamPluginsManager.ts
```diff
@@ -2,19 +2,38 @@ import { HMSMediaStreamPlugin } from './HMSMediaStreamPlugin';
...
+              throw Error(errorMessage);
+            } else {
+              HMSLogger.w(this.TAG, errorMessage);
+            }
+          }
+          break;
+        default:
+      }
+      this.plugins.add(plugin);
```
raviteja83, line 35: this would still execute right? in-case of warning?

KaustubhKumar05: Ack, will add a continue. Was unhandled since I was throwing an error earlier

raviteja83 (approved):

PR #2948, opened 2024-05-28
feat: add `findPeerByName` API for large rooms

raviteja83:

-
-


packages/hms-video-store/src/signal/interfaces/superpowers.ts
```diff
@@ -95,6 +95,12 @@ export interface findPeersRequestParams {
   limit: number;
 }
 
+export interface findPeerByNameRequestParams {
```
eswarclynn, line 98: Should these interface names be capitalized?

eswarclynn (approved):

PR #2987, opened 2024-06-12
feat: send krisp audio analytics detail

amar-1995:

-
-


packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -243,6 +243,19 @@ export default class AnalyticsEventFactory {
     });
   }
 
+  static getKrispStart() {
```
raviteja83, line 246: get prefix not needed for these two.

raviteja83 (approved):

PR #2988, opened 2024-06-12
fix: stop caption position rerendering

amar-1995:


packages/roomkit-react/src/Prebuilt/layouts/VideoStreamingSection.tsx
```diff
@@ -140,7 +138,7 @@ export const VideoStreamingSection = ({
             // @ts-ignore
             return <GridLayout {...(elements as DefaultConferencingScreen_Elements)?.video_tile_layout?.grid} />;
           })}
-        <CaptionsViewer />
+        <CaptionsViewer setDefaultPosition={setCaptionPosition} defaultPosition={captionPosition} />
```
raviteja83, line 141: this can be just position and setPosition as it updates to current postion and not a default.

raviteja83 (approved):

PR #2989, opened 2024-06-12
fix: caption ui enhancements

amar-1995:


packages/roomkit-react/src/Prebuilt/components/MoreSettings/CaptionContent.tsx
```diff
@@ -106,6 +107,7 @@ export const CaptionContent = ({ isMobile, onExit }: { isMobile: boolean; onExit
                 await actions.startTranscription({
                   mode: HMSTranscriptionMode.CAPTION,
                 });
+                setIsCaptionEnabled(true);
```
raviteja83, line 110: start is only successful if admin?

raviteja83 (approved):

PR #2995, opened 2024-06-13
refactor: noise cancellation component, add to mweb when supported

amar-1995:


packages/roomkit-react/src/Prebuilt/components/Preview/PreviewJoin.tsx
```diff
@@ -264,7 +272,14 @@ export const PreviewControls = ({ hideSettings, vbEnabled }: { hideSettings: boo
         <AudioVideoToggle />
         {vbEnabled ? <VBToggle /> : null}
       </Flex>
-      {!hideSettings ? <PreviewSettings /> : null}
+      <Flex align="center" gap="1">
+        {isMobile && isNoiseCancellationEnabled && isKrispPluginAdded ? (
```
raviteja83, line 276: it can be shown otherwise also no? why are we showing only when added? if it is not enabled by default, they should be able to enable.

raviteja83, line 276: we should krispPlugin.checkSupport.isSupported as welll.

packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/MwebOptions.tsx
```diff
@@ -197,6 +209,19 @@ export const MwebOptions = ({
                 <ActionTile.Title>{isHandRaised ? 'Lower' : 'Raise'} Hand</ActionTile.Title>
               </ActionTile.Root>
             ) : null}
+            {krispPlugin.isSupported() && room.isNoiseCancellationEnabled && localPeerAudioTrackID ? (
```
raviteja83, line 212: why is this needed? we can;t use the noisecancellation from AudioVideoToggle? it does have isMobile check no?

packages/roomkit-react/src/Prebuilt/components/Preview/PreviewJoin.tsx
amar-1995: This is only an icon in the preview screen that says whether noise cancellation is enabled.

packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/MwebOptions.tsx
amar-1995: We need to show it more options in mweb

raviteja83 (approved):

PR #3010, opened 2024-06-18
feat: lazy load less frequently used components 

raviteja83:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
+++ b/packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
@@ -286,7 +286,7 @@ const HLSView = () => {
         hlsPlayer.reset();
       };
     }
-  }, [hlsUrl, vanillaStore, hmsActions]);
+  }, [hlsUrl, vanillaStore, hmsActions, hoverControlsVisible]);
 
   /**
    * initialize and subscribe to hlsState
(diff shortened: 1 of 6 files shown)


scripts/build-webapp.js
```diff
@@ -16,7 +16,7 @@ async function main() {
   const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
   const source = pkg.name === '@100mslive/roomkit-web' ? './src/index.js' : './src/index.ts';
   const external = [...Object.keys(pkg.dependencies || {}), ...Object.keys(pkg.peerDependencies || {})];
-  const loader = { '.js': 'jsx', '.svg': 'dataurl', '.png': 'dataurl' };
+  const loader = { '.js': 'jsx', '.svg': 'file', '.png': 'file' };
```
KaustubhKumar05, line 19: Why was this changed?

raviteja83: inlining is increasing the js size. so we will load externally. it can be cached in cdn as well. Need to check the improvement here.

packages/roomkit-react/src/Prebuilt/layouts/HLSView.jsx
```diff
@@ -286,7 +286,7 @@ const HLSView = () => {
         hlsPlayer.reset();
       };
     }
-  }, [hlsUrl, vanillaStore, hmsActions]);
+  }, [hlsUrl, vanillaStore, hmsActions, hoverControlsVisible]);
```
amar-1995, line 289: Just once check if it won't reinitialise the hls player again.

KaustubhKumar05 (dismissed):

KaustubhKumar05:
We should probably do a round of regression testing this release since we might miss some flows during dev test

amar-1995 (commented):
We need to do one load test, it will help us to analysis the improvement from user perspective. cc: @sahil-100ms

hdz-666 (dismissed):

KaustubhKumar05 (approved):

ygit (approved):

PR #3013, opened 2024-06-19
fix: pre call diagnostics ui issues

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/diagnostics/interfaces.ts
+++ b/packages/hms-video-store/src/diagnostics/interfaces.ts
@@ -27,6 +27,7 @@ export interface MediaPermissionCheck {
 
 export interface HMSDiagnosticsInterface {
   requestPermission(check: MediaPermissionCheck): Promise<MediaPermissionCheck>;
+  validateBrowserAPIs(): void;
   startMicCheck(inputDevice?: string, onStop?: () => void, time?: number): Promise<void>;
   getRecordedAudio(): string | undefined;
   stopMicCheck(): void;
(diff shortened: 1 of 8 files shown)


packages/hms-video-store/src/diagnostics/interfaces.ts
```diff
@@ -27,6 +27,7 @@ export interface MediaPermissionCheck {
 
 export interface HMSDiagnosticsInterface {
   requestPermission(check: MediaPermissionCheck): Promise<MediaPermissionCheck>;
+  validateBrowserAPIs(): void;
```
raviteja83, line 30: checkBrowserSupport

raviteja83: we can comment on each explaining what it does.

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
```diff
@@ -76,15 +81,30 @@ const DiagnosticsStepsList = () => {
...
-            </Text>
-          </li>
-        ))}
-      </ul>
+      {Object.keys(DiagnosticsSteps).map(key => {
+        const isStepCompleted =
+          Object.keys(DiagnosticsSteps).findIndex(step => step === activeStep) >
+          Object.keys(DiagnosticsSteps).findIndex(step => step === key);
+
+        let c = '$on_primary_low';
```
raviteja83, line 89: use color instead.

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
```diff
@@ -76,15 +81,30 @@ const DiagnosticsStepsList = () => {
...
-          <li key={key}>
-            <Text variant="md" css={{ mb: '$10', c: activeStep === key ? '$on_primary_high' : '$on_primary_low' }}>
-              {DiagnosticsSteps[key]}
-            </Text>
-          </li>
-        ))}
-      </ul>
+      {Object.keys(DiagnosticsSteps).map(key => {
+        const isStepCompleted =
+          Object.keys(DiagnosticsSteps).findIndex(step => step === activeStep) >
```
raviteja83, line 86: can we use one iteration. no easy way other than this?

raviteja83: can't we pass a prop or something? why are we again finding active step?

raviteja83 (approved):

PR #3015, opened 2024-06-19
fix: use phone icon for earpiece on mweb

amar-1995:


packages/roomkit-react/src/Prebuilt/components/Header/common.jsx
```diff
@@ -82,6 +83,8 @@ export const AudioActions = () => {
     AudioIcon = <BluetoothIcon />;
   } else if (currentSelection && currentSelection.label.toLowerCase().includes('wired')) {
     AudioIcon = <HeadphonesIcon />;
+  } else if (currentSelection && currentSelection.label.toLowerCase().includes('headset')) {
```
raviteja83, line 86: should be earpiece not headset

raviteja83 (approved):

PR #3025, opened 2024-06-20
fix: completion state for connectivity test, audio test ui

KaustubhKumar05:
### Fixes:

https://100ms.atlassian.net/browse/WEB-2884 (Connection quality completion state)
https://100ms.atlassian.net/browse/WEB-2887 (Output selection for Safari)
https://100ms.atlassian.net/browse/WEB-2891 (Update audio test UI)
https://100ms.atlassian.net/browse/WEB-2893 (Output selection for mweb)

### Before:

image

image

### After:

Going back to in progress if the test is run again and on completion, the state is updated

image

### DIFF
--- a/packages/roomkit-react/src/Diagnostics/AudioTest.tsx
+++ b/packages/roomkit-react/src/Diagnostics/AudioTest.tsx
@@ -1,6 +1,5 @@
 /* eslint-disable react/prop-types */
 import React, { useEffect, useState } from 'react';
-import { useMedia } from 'react-use';
 import {
   selectDevices,
   selectLocalAudioTrackID,
@@ -15,12 +14,11 @@ import { Button } from '../Button';
 import { Box, Flex } from '../Layout';
 import { Progress } from '../Progress';
 import { Text } from '../Text';
-import { config as cssConfig } from '../Theme';
 // @ts-ignore: No implicit any
 import { DeviceSelector } from './DeviceSelector';
 import { hmsDiagnostics } from './hms';
 import { useAudioOutputTest } from '../Prebuilt/components/hooks/useAudioOutputTest';
-import { isSafari, TEST_AUDIO_URL } from '../Prebuilt/common/constants';
+import { TEST_AUDIO_URL } from '../Prebuilt/common/constants';
 
 const SelectContainer = ({ children }: { children: React.ReactNode }) => (
   <Box css={{ w: '50%', '@lg': { w: '100%' } }}>{children}</Box>
@@ -112,9 +110,8 @@ const SpeakerTest = () => {
   const actions = useHMSActions();
   const devices = useHMSStore(selectDevices);
   const { audioOutputDeviceId } = useHMSStore(selectLocalMediaSettings);
-  const isMobile = useMedia(cssConfig.media.md);
 
-  if (isMobile || isSafari) {
+  if (devices.audioOutput.length === 0) {
     return <></>;
   }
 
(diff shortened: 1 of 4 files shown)


packages/roomkit-react/src/Diagnostics/AudioTest.tsx
```diff
@@ -84,37 +112,23 @@ const SpeakerTest = () => {
   const actions = useHMSActions();
   const devices = useHMSStore(selectDevices);
   const { audioOutputDeviceId } = useHMSStore(selectLocalMediaSettings);
-  const { playing, setPlaying, audioRef } = useAudioOutputTest({ deviceId: audioOutputDeviceId || 'default' });
+  const isMobile = useMedia(cssConfig.media.md);
+
+  if (isMobile || isSafari) {
```
raviteja83, line 117: in safari do we get the output device in device list? we can check the length of audioOutput. That would be a better check instead of isMobile and isSafari. covers both.

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
```diff
@@ -77,23 +77,20 @@ const DiagnosticsStep = () => {
...
+  const { activeStep, connectivityTested } = useContext(DiagnosticsContext);
 
   return (
     <Box css={{ w: '25%', '@lg': { display: 'none' } }}>
       {Object.keys(DiagnosticsSteps).map(key => {
         const keys = Object.keys(DiagnosticsSteps);
         const activeStepIndex = keys.indexOf(activeStep);
         const keyIndex = keys.indexOf(key);
-        const isStepCompleted = activeStepIndex > keyIndex;
+        const isStepCompleted = activeStepIndex > keyIndex || (activeStep === 'connectivity' && connectivityTested);
```
raviteja83, line 88: possible to have a state in the context.

[{ 
  step: 'name',
 completed: boolean,
active: boolean
},
{
step: 'name',
 completed: boolean,
active: boolean
}

]

packages/roomkit-react/src/Diagnostics/AudioTest.tsx
KaustubhKumar05: Ack, will check this

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
KaustubhKumar05: This is only needed for the final step since it can be re-run. For the others the existing logic of checking the active step index is sufficient, didn't want to complicate it. Can update if this has some edge cases

eswarclynn: This what what I was thinking earlier, we also need an error property here. I can pick this up next

packages/roomkit-react/src/Diagnostics/AudioTest.tsx
KaustubhKumar05: This works, updated the PR

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
raviteja83: Let us keep it as enahancements then.

raviteja83 (approved):

packages/roomkit-react/src/Diagnostics/Diagnostics.tsx
```diff
@@ -77,23 +77,20 @@ const DiagnosticsStep = () => {
...
 
         let color = '$on_primary_low';
         if (isStepCompleted) {
           color = '$primary_bright';
         }
-        if (activeStep === key) {
-          color = '$on_primary_high';
-        }
```
eswarclynn, line 96: Why was this removed @KaustubhKumar05?

PR #3031, opened 2024-06-21
fix: audio test recording with audio silence error, permission error modal

eswarclynn:

-
-


packages/hms-video-store/src/diagnostics/interfaces.ts
```diff
@@ -28,7 +28,12 @@ export interface MediaPermissionCheck {
 export interface HMSDiagnosticsInterface {
   requestPermission(check: MediaPermissionCheck): Promise<MediaPermissionCheck>;
   checkBrowserSupport(): void;
-  startMicCheck(inputDevice?: string, onStop?: () => void, time?: number): Promise<void>;
+  startMicCheck(
```
raviteja83, line 31: better to take an object, easy to pass and understood by key

PR #3060, opened 2024-07-03
feat: added default audio in screenshare

amar-1995:

-
-

### DIFF
--- a/packages/hms-video-store/src/sdk/LocalTrackManager.ts
+++ b/packages/hms-video-store/src/sdk/LocalTrackManager.ts
@@ -219,7 +219,10 @@ export class LocalTrackManager {
         echoCancellation: false,
       };
     } else {
-      constraints.audio = true;
+      constraints.audio = {
+        noiseSuppression: false,
+        echoCancellation: false,
+      };
     }
     let stream;
     try {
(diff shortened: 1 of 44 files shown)


packages/hms-video-store/src/sdk/LocalTrackManager.ts
```diff
@@ -218,6 +218,8 @@ export class LocalTrackManager {
         googAutoGainControl: false,
         echoCancellation: false,
       };
+    } else {
+      constraints.audio = true;
```
raviteja83, line 222: instead of true it should be an object like above disabling noiseSuppression, echoCancellation(not sure if this would result in echo)

packages/hms-video-store/src/sdk/LocalTrackManager.ts
```diff
@@ -218,6 +218,11 @@ export class LocalTrackManager {
         googAutoGainControl: false,
         echoCancellation: false,
       };
+    } else {
+      constraints.audio = {
```
raviteja83, line 222: should be set only when tab sharing is done. let us figure out. does it work on all browsers as is?

raviteja83 (approved):

PR #3068, opened 2024-07-08
feat: retry subscribe connection with relay

raviteja83:

-
-


packages/hms-video-store/src/transport/index.ts
```diff
@@ -1122,10 +1124,19 @@ export default class HMSTransport {
...
       });
 
-      return Promise.race([p, timeout]) as Promise<boolean>;
+      return Promise.race([p, timeout]).then(value => {
+        // value is false if the timeout got resolved before the subscribe connection
+        if (!value) {
+          // close and reset subscribe connection to retry with relay
+          this.subscribeConnection?.close();
+          this.subscribeConnection = null;
+          this.createPeerConnections('relay');
```
adityaa30, line 1136: I don't think we create PeerConnection again OR overwrite the previous one. During IceRestart the PeerConnection instance should be re-used as it contains metadata about the ice candidates (trickle) and tracks (previous offer/answer SDP). 

Not sure if we can change the `transportPolicy` on the fly.

raviteja83: as of now it is only on the configuration. I don't see any api to update the transport policy

eswarclynn (approved):

PR #3070, opened 2024-07-08
fix: role change not publishing/unpublishing when happened during offline -> online

raviteja83:

-
-

### DIFF
--- a/packages/hms-video-store/src/transport/index.ts
+++ b/packages/hms-video-store/src/transport/index.ts
@@ -1,3 +1,4 @@
+import EventEmitter2 from 'eventemitter2';
 import { JoinParameters } from './models/JoinParameters';
 import { TransportFailureCategory } from './models/TransportFailureCategory';
 import { TransportState } from './models/TransportState';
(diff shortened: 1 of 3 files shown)


packages/hms-video-store/src/transport/index.ts
```diff
@@ -619,46 +601,31 @@ export default class HMSTransport {
...
-            `Setting maxBitrate=${track.settings.maxBitrate} kpbs${
-              track instanceof HMSLocalVideoTrack ? ` and maxFramerate=${track.settings.maxFramerate}` : ''
-            } for ${track.source} ${track.type} ${track.trackId}`,
-          );
-        })
-        .catch(error => HMSLogger.w(TAG, 'Failed setting maxBitrate and maxFramerate', error));
-
-      track.isPublished = true;
+    // @ts-ignore
+    await this.eventEmitter.waitFor(RENEGOTIATION_CALLBACK_ID, { handleError: true });
```
eswarclynn, line 610: How is handleError of use here? From the event emitter docs I can see that it's used only in the return value of the promise? https://github.com/EventEmitter2/EventEmitter2?tab=readme-ov-file#emitterwaitforevent--eventns-filter

raviteja83: when handleError is used, it will throw if the first arg is of type error.
You can locally test by throwing an error.

eswarclynn (approved):

PR #3083, opened 2024-07-11
fix: whiteboard reconnection

eswarclynn:

-
-


packages/hms-whiteboard/src/hooks/StoreClient.ts
```diff
@@ -72,10 +72,8 @@ export class SessionStore<T> {
       }
     });
 
-    count = await this.getKeysCount();
-    if (!count) {
-      handleOpen([]);
-    }
+    count = await this.getKeysCountWithDelay();
```
raviteja83, line 75: this will become unhandled error though.

raviteja83: are all cases handled here?

raviteja83 (approved):

PR #3084, opened 2024-07-12
feat: switch role for local peer

amar-1995:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
@@ -369,7 +369,6 @@ const ParticipantMoreActions = ({ peerId, role }: { peerId: string; role: string
     isInStage,
     shouldShowStageRoleChange,
   } = usePeerOnStageActions({ peerId, role });
-  const localPeerId = useHMSStore(selectLocalPeerID);
   const canChangeRole = !!useHMSStore(selectPermissions)?.changeRole;
   const [openRoleChangeModal, setOpenRoleChangeModal] = useState(false);
   const roles = useHMSStore(selectAvailableRoleNames);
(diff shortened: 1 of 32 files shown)


packages/roomkit-react/src/Prebuilt/components/TileMenu/TileMenuContent.tsx
```diff
@@ -281,6 +281,20 @@ export const TileMenuContent = ({
             </Text>
           </StyledMenuTile.ItemButton>
         )}
+        {canChangeRole && roles.length > 1 ? (
```
raviteja83, line 284: we discussed it should be in more options no? why did we add it here again?

raviteja83: @amar-1995 can we remove from tile menu?

raviteja83: or if it is needed please share the thread internally

packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
```diff
@@ -403,7 +401,7 @@ const ParticipantMoreActions = ({ peerId, role }: { peerId: string; role: string
         </Dropdown.Trigger>
         <Dropdown.Portal>
           <Dropdown.Content align="end" sideOffset={8} css={{ w: '$64', bg: '$surface_default' }}>
-            {shouldShowStageRoleChange ? (
+            {shouldShowStageRoleChange && localPeerId !== peerId ? (
```
raviteja83, line 404: this check might be handled from customiser. mostly not required though. if role change is allowed we can allow to get back from stage i think

raviteja83 (approved):

PR #3085, opened 2024-07-12
feat: add processed audio track only when plugin is added

raviteja83:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
@@ -109,6 +109,9 @@ const useNoiseCancellationWithPlugin = () => {
     if (inProgress) {
       return;
     }
+    if (!krispPlugin.checkSupport().isSupported) {
+      throw Error('Krisp plugin is not supported');
+    }
     setInProgress(true);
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
```diff
@@ -281,13 +284,17 @@ export const AudioVideoToggle = ({ hideOptions = false }: { hideOptions?: boolea
...
       if (isNoiseCancellationEnabled && !isKrispPluginAdded && !inProgress && localPeer?.audioTrack) {
-        await setNoiseCancellationWithPlugin(true);
-        ToastManager.addToast({
-          title: `Noise Reduction Enabled`,
-          variant: 'standard',
-          duration: 2000,
-          icon: <AudioLevelIcon />,
-        });
+        try {
+          await setNoiseCancellationWithPlugin(true);
```
KaustubhKumar05, line 288: Should this be set to false inside the catch block?

raviteja83: plugin won't be added, no action needed

KaustubhKumar05 (approved):

PR #3087, opened 2024-07-15
fix: active speaker font size

KaustubhKumar05:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Header/HeaderComponents.jsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Header/HeaderComponents.jsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { selectDominantSpeaker, useHMSStore } from '@100mslive/react-sdk';
 import { VolumeOneIcon } from '@100mslive/react-icons';
-import { Flex, styled, Text, textEllipsis, VerticalDivider } from '../../../';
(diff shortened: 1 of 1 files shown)


packages/roomkit-react/src/Prebuilt/components/Header/HeaderComponents.jsx
```diff
@@ -14,9 +14,8 @@ export const SpeakerTag = () => {
         justify="center"
         css={{ flex: '1 1 0', color: '$on_surface_high', '@md': { display: 'none' } }}
       >
-        <VerticalDivider css={{ ml: '$8' }} />
         <VolumeOneIcon />
-        <Text variant="md" css={{ ...textEllipsis(200), ml: '$2' }} title={dominantSpeaker.name}>
+        <Text variant="sm" css={{ ...textEllipsis(200), ml: '$2' }} title={dominantSpeaker.name}>
```
raviteja83, line 18: wasn't the issue typography? was it only the size?

KaustubhKumar05: https://www.figma.com/design/ELM8oi6Abp9w1dzF5ApiJA/Prebuilt-Kit---Stage-1.5?node-id=7341-158641&t=YfTOwG4qMc7x8Bgc-0
The design was shared later, only the font size had to be updated

amar-1995 (approved):

PR #3093, opened 2024-07-18
fix: pop out chat fixes

KaustubhKumar05:
Close original chat window when opening pop out chat  
Pop out chat should be open with latest message in view  
Add “To XYZ” for DMs  
Align textarea placeholder vertically  
Overscroll shows white background  
Verify block and paused states
https://100ms.atlassian.net/browse/WEB-2935

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import {
   selectHMSMessages,
   selectLocalPeerID,
+  selectPeerNameByID,
   selectSessionStore,
   selectUnreadHMSMessagesCount,
   useHMSStore,
@@ -14,6 +15,7 @@ import { Tooltip } from '../../../Tooltip';
 import IconButton from '../../IconButton';
 import { AnnotisedMessage } from '../Chat/ChatBody';
 import { useRoomLayoutConferencingScreen } from '../../provider/roomLayoutProvider/hooks/useRoomLayoutScreen';
+import { useIsPeerBlacklisted } from '../hooks/useChatBlacklist';
 import { CHAT_MESSAGE_LIMIT, formatTime } from '../Chat/utils';
 import { SESSION_STORE_KEY } from '../../common/constants';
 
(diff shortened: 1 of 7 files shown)


packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
```diff
@@ -55,7 +60,7 @@ export const PIPChat = () => {
           bg: '$surface_dim',
           overflowY: 'auto',
           // Subtracting height of footer
-          h: canSendChatMessages ? 'calc(100% - 90px)' : '100%',
+          h: canSendChatMessages ? 'calc(100% - 87px)' : '100%',
```
raviteja83, line 63: fixed pixel can break at somepoint no?

packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
```diff
@@ -193,20 +193,36 @@ export const PIPChat = () => {
...
               id="chat-input"
               maxLength={CHAT_MESSAGE_LIMIT}
+              disabled={!isChatEnabled || !isLocalPeerBlacklisted}
+              rows={1}
               style={{ border: 'none', resize: 'none' }}
               css={{
                 w: '100%',
                 c: '$on_surface_high',
-                padding: '0.25rem !important',
+                p: '0.75rem 0.75rem !important',
```
raviteja83, line 202: why are we using both style and css props? ca we use one

packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
```diff
@@ -193,20 +193,36 @@ export const PIPChat = () => {
...
               }}
-              placeholder={message_placeholder}
+              placeholder={
+                isLocalPeerBlacklisted
```
raviteja83, line 205: better to have a fn outside the UI to gt text. not readable.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPWindow.tsx
```diff
@@ -9,5 +9,6 @@ type PIPWindowProps = {
 export const PIPWindow = ({ pipWindow, children }: PIPWindowProps) => {
   pipWindow.document.body.style.margin = '0';
   pipWindow.document.body.style.overflowX = 'hidden';
+  pipWindow.document.body.style.overflowY = 'clip';
```
raviteja83, line 12: we can clip overflow no? do we need two different overflow types?

packages/roomkit-react/src/Prebuilt/components/PIP/PIPChat.tsx
KaustubhKumar05: The chat container is flexible dimensions, the footer has fixed height and flexible width. This just leaves some height for the footer and grows otherwise

raviteja83 (approved):

amar-1995 (approved):

PR #3095, opened 2024-07-18
fix: event handlers initialised twice for pop out chat

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/components/Notifications/Notifications.tsx
```diff
@@ -53,7 +53,7 @@ export function Notifications() {
   const { showNotification } = useAwayNotifications();
   const amIScreenSharing = useHMSStore(selectIsLocalScreenShared);
   const logoURL = useRoomLayout()?.logo?.url;
-  const { pipWindow } = usePIPChat();
```
KaustubhKumar05, line 56: usePIPChat internally calls usePIPWindow and adds event listeners. Instead directly using usePIPWindow

amar-1995 (approved):

raviteja83 (approved):

PR #3106, opened 2024-07-22
fix: move testid to buttons

KaustubhKumar05:
Before:

image

After:

image

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
@@ -310,10 +310,9 @@ export const AudioVideoToggle = ({ hideOptions = false }: { hideOptions?: boolea
           disabled={!toggleAudio}
           hideOptions={hideOptions || !hasAudioDevices}
           onDisabledClick={toggleAudio}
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/components/AudioVideoToggle.tsx
```diff
@@ -310,10 +310,9 @@ export const AudioVideoToggle = ({ hideOptions = false }: { hideOptions?: boolea
           disabled={!toggleAudio}
           hideOptions={hideOptions || !hasAudioDevices}
           onDisabledClick={toggleAudio}
+          testid={!isLocalAudioEnabled ? 'audio_off_btn' : 'audio_on_btn'}
```
raviteja83, line 313: can be just audio_toggle_btn

KaustubhKumar05: Will check with Sahil, he might have to update the tests in this case

raviteja83: right now i don't think these id/s are used right? it might be used differently on tests

eswarclynn (approved):

amar-1995 (approved):

PR #3122, opened 2024-07-24
feat: interruption events

amar-1995:

-
-


packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -76,6 +76,16 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
   private handleVisibilityChange = async () => {
     if (document.visibilityState === 'visible') {
       await this.replaceTrackWith(this.settings);
+      // stopped
+      AnalyticsEventFactory.interrupionStop(this.type, {
```
raviteja83, line 80: add a conditional for the fn and then pass the params. args are same

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -60,7 +60,7 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
     }
     this.pluginsManager = new HMSAudioPluginsManager(this, eventBus, room);
     this.setFirstTrackId(track.id);
-    if (isIOS() && isBrowser) {
+    if (isMobile() && isBrowser) {
```
raviteja83, line 63: this works on android. why are we changing without an understanding? this will increase testing scope

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -256,6 +257,28 @@ export default class AnalyticsEventFactory {
...
+      name: 'interruption.start',
+      level: AnalyticsEventLevel.INFO,
+      properties: {
+        type,
+        ...deviceInfo,
+      },
+    });
+  }
+
+  static interrupionStop(type: HMSTrackType, deviceInfo: Partial<MediaDeviceInfo>) {
```
raviteja83, line 272: one event is enough. you can take a bool for start or stop and add a condition in name. both are exactly same

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
amar-1995: I tested it on android, it is not working. Condition above is checking for ios browser only.

raviteja83:
AudioSinkManager also has interuption that also has to be handled.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
raviteja83: does the change make a difference?

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
raviteja83: only mark as resolved after you push the changes.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
amar-1995: I added this check which I not exactly remember. But it is only for ios device where we want to reacquire the audio.
https://github.com/100mslive/web-sdks/pull/1113/files

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -7,6 +7,7 @@ import { HMSException } from '../error/HMSException';
 import { DeviceMap, SelectedDevices } from '../interfaces';
 import { HMSTrackSettings } from '../media/settings/HMSTrackSettings';
 import { HMSRemoteVideoTrack } from '../media/tracks/HMSRemoteVideoTrack';
+import { HMSTrackType } from '../schema';
```
raviteja83, line 10: i don't think should be from schema. see if the same is there elsewhere.

packages/hms-video-store/src/media/tracks/HMSTrack.ts
```diff
@@ -84,7 +85,16 @@ export abstract class HMSTrack {
   protected setFirstTrackId(trackId: string) {
     this.firstTrackId = trackId;
   }
-
+  /**
+   * @internal
+   * It will send event to analytics when interruption start/stop
+   */
+  sendInterruptionEvent(started: boolean, isRemoteAudio = false) {
+    AnalyticsEventFactory.interrupion(started, isRemoteAudio ? 'remote.audio' : this.type, {
```
raviteja83, line 93: this will only give the event. where are we sending it?

packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -116,13 +116,22 @@ export class AudioSinkManager {
...
+      audioTrack.sendInterruptionEvent(true, true);
     }
   };
 
+  private handleAudioPlay = async (event: any) => {
+    // this means the audio replayed because of external factors(headset removal, incoming phone call)
+    const audioTrack = this.store.getTrackById(event.target.id);
+    if (audioTrack) {
+      // stopped interruption event
+      audioTrack.sendInterruptionEvent(false, true);
```
raviteja83, line 132: we should probably send only once, not for each track. will be too many events in a large room.

packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -140,6 +149,7 @@ export class AudioSinkManager {
     audioEl.style.display = 'none';
     audioEl.id = track.trackId;
     audioEl.addEventListener('pause', this.handleAudioPaused);
+    audioEl.addEventListener('play', this.handleAudioPlay);
```
raviteja83, line 152: don't this this helps much. we can mark as false when the track is played again. there is a method to replay.

packages/hms-video-store/src/media/tracks/HMSTrack.ts
```diff
@@ -84,7 +85,16 @@ export abstract class HMSTrack {
   protected setFirstTrackId(trackId: string) {
     this.firstTrackId = trackId;
   }
-
+  /**
+   * @internal
+   * It will send event to analytics when interruption start/stop
+   */
+  sendInterruptionEvent({ started, isRemoteAudio = false }: { started: boolean; isRemoteAudio?: boolean }) {
+    return AnalyticsEventFactory.interrupion(started, isRemoteAudio ? 'remote.audio' : this.type, {
```
raviteja83, line 93: this is not sent to the server

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -74,8 +77,22 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
   }
 
   private handleVisibilityChange = async () => {
-    if (document.visibilityState === 'visible') {
+    if (this.nativeTrack.readyState === 'ended' && this.nativeTrack.muted && document.visibilityState !== 'visible') {
```
raviteja83, line 80: this is not necessarily true. it should be ended or muted

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -74,8 +77,22 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
+      this.eventBus.analytics.publish(
+        this.sendInterruptionEvent({
+          started: true,
+        }),
+      );
+      this.isInterrupted = true;
```
raviteja83, line 86: why this extra flag, add unnecessary complexity, the track status can be checked when visibile

packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -499,9 +499,21 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       this.enabledStateBeforeBackground = this.enabled;
       this.nativeTrack.enabled = false;
       this.replaceSenderTrack(this.nativeTrack);
```
raviteja83, line 501: can we check if this is needed. not relveant to this PR but good to check if it is needed

packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -499,9 +499,21 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       this.enabledStateBeforeBackground = this.enabled;
```
raviteja83, line 499: add an early return when source !== 'regular' otherwise for screenshare also it might go into the else.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -72,10 +69,22 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
   resetManuallySelectedDeviceId() {
     this.manuallySelectedDeviceId = undefined;
   }
-
+  // eslint-disable-next-line complexity
   private handleVisibilityChange = async () => {
-    if (document.visibilityState === 'visible') {
+    if ((this.nativeTrack.readyState === 'ended' || this.nativeTrack.muted) && document.visibilityState !== 'visible') {
```
raviteja83, line 74: can we add the same check here for regular that is added to video.

packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -510,7 +510,7 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
       );
     } else {
       this.nativeTrack.enabled = this.enabledStateBeforeBackground;
-      this.replaceSenderTrack(this.processedTrack || this.nativeTrack);
+      // this.replaceSenderTrack(this.processedTrack || this.nativeTrack);
```
raviteja83, line 513: why is this removed?

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -73,9 +72,30 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
   private handleVisibilityChange = async () => {
-    if (document.visibilityState === 'visible') {
+    // track state is fine do nothing
+    if (!this.isTrackNotPublishing()) {
+      HMSLogger.d(this.TAG, `visibiltiy: ${document.visibilityState}`, `${this}`);
+      return;
+    }
+    if (document.visibilityState === 'hidden') {
+      this.eventBus.analytics.publish(
+        this.sendInterruptionEvent({
```
KaustubhKumar05, line 87: We were not getting interruption events for audio track yesterday, is that working now?

raviteja83: you were checking in audio sink manager I think. it might work not so sure. have not tested myself.

KaustubhKumar05: Was checking for events on events inspector as well

amar-1995: @raviteja83 Audio is not working because the track state is not ended or muted.

amar-1995: I tried going into the background. I tried calling as well with @KaustubhKumar05 on Monday. It was not working

raviteja83: might work on iOS.

raviteja83 (approved):

eswarclynn (approved):

PR #3124, opened 2024-07-25
fix: pick bt device on first load, select non-default version of default device

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/device-manager/DeviceManager.ts
+++ b/packages/hms-video-store/src/device-manager/DeviceManager.ts
@@ -434,7 +434,7 @@ export class DeviceManager implements HMSDeviceManager {
         speakerPhone = device;
       } else if (label.includes('wired')) {
         wired = device;
-      } else if (/airpods|buds|bluetooth/.test(label)) {
+      } else if (/airpods|buds|wireless|bluetooth/gi.test(label)) {
         bluetoothDevice = device;
       } else if (label.includes('earpiece')) {
         earpiece = device;
--- a/packages/hms-video-store/src/device-manager/DeviceManager.ts
+++ b/packages/hms-video-store/src/device-manager/DeviceManager.ts
@@ -206,7 +206,7 @@ export class DeviceManager implements HMSDeviceManager {
     const audioDeviceId = this.store.getConfig()?.settings?.audioInputDeviceId;
     if (!audioDeviceId && localPeer?.audioTrack) {
       const getInitialDeviceId = () => {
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/device-manager/DeviceManager.ts
```diff
@@ -205,7 +205,12 @@ export class DeviceManager implements HMSDeviceManager {
     }
     const audioDeviceId = this.store.getConfig()?.settings?.audioInputDeviceId;
     if (!audioDeviceId && localPeer?.audioTrack) {
-      await localPeer.audioTrack.setSettings({ deviceId: this.audioInput[0]?.deviceId }, true);
+      const getInitialDeviceId = () => {
+        const isMobile = !('setSinkId' in HTMLMediaElement.prototype);
```
raviteja83, line 209: this willl be true for firefox and safari desktop I think.

raviteja83: as discussed, if verified on android, we can test with just isIOS

packages/hms-video-store/src/device-manager/DeviceManager.ts
```diff
@@ -416,6 +421,29 @@ export class DeviceManager implements HMSDeviceManager {
...
+    let wired: InputDeviceInfo | null = null;
+    let earpiece: InputDeviceInfo | null = null;
+
+    for (const device of this.audioInput) {
+      const label = device.label.toLowerCase();
+      if (label.includes('speakerphone')) {
+        speakerPhone = device;
+      } else if (label.includes('wired')) {
+        wired = device;
+      } else if (/airpods|buds|bluetooth/.test(label)) {
```
raviteja83, line 437: add gi at the end of regex. pro can also be one that might be common.

raviteja83: wireless can also be added.

eswarclynn: label is already converted to lowercase, so 'i' in the regex won't be required?

Will add wireless but how do you think pro is useful, our MacBook Pros can also be misidentified as Bluetooth devices in this case?

raviteja83: yeah pro we can ignore. good point.

raviteja83 (approved):

KaustubhKumar05 (approved):

PR #3127, opened 2024-07-28
fix: rate limit events

KaustubhKumar05:
- [x] Update limit based on normal usage


packages/hms-video-store/src/analytics/AnalyticsTransport.ts
```diff
@@ -9,12 +9,31 @@ export abstract class AnalyticsTransport {
...
   private readonly TAG = '[AnalyticsTransport]';
 
+  private eventCount = 0;
+  private lastResetTime: number = Date.now();
+  private readonly MAX_EVENTS_PER_MINUTE: number = 200;
+  private readonly RESET_INTERVAL_MS: number = 60000;
+
+  private checkRateLimit() {
+    const now = Date.now();
+    if (now - this.lastResetTime >= this.RESET_INTERVAL_MS) {
```
raviteja83, line 19: should we make it specific to event? sometimes large rooms might send more events not sure. ideally should not be the case.

amar-1995: Yes, It will depend on number of peers in a room.
But here we are talking about local peer receiving 200 events/min, That should be enough for now. We could have analytics to send max event reached as well. We will make alert on max-events reached.
Later as per of requirement we can check and make it event based.

eswarclynn (approved):

packages/hms-video-store/src/analytics/AnalyticsTransport.ts
```diff
@@ -9,12 +9,31 @@ export abstract class AnalyticsTransport {
...
+  }
+
   sendEvent(event: AnalyticsEvent) {
     try {
+      this.checkRateLimit();
       this.sendSingleEvent(event);
       this.flushFailedEvents();
     } catch (error) {
       HMSLogger.w(this.TAG, 'sendEvent failed', error);
+      throw error;
```
eswarclynn, line 36: Should we throw error if sending analytics fails @raviteja83? I think we don't need it and that's why there's just a warning here.

KaustubhKumar05: Updated

PR #3148, opened 2024-08-02
fix: update error, add node id to offer/answer/trickle

raviteja83:

-
-

### DIFF
--- a/packages/hms-video-store/src/transport/index.ts
+++ b/packages/hms-video-store/src/transport/index.ts
@@ -420,6 +420,7 @@ export default class HMSTransport {
   }
 
   setSFUNodeId(id: string) {
+    this.signal.setSfuNodeId(id);
     if (!this.sfuNodeId) {
       this.sfuNodeId = id;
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/transport/index.ts
```diff
@@ -420,6 +420,7 @@ export default class HMSTransport {
   }
 
   setSFUNodeId(id: string) {
+    this.signal.setSfuNodeId(id);
     if (!this.sfuNodeId) {
       this.sfuNodeId = id;
```
KaustubhKumar05, line 425: This can be moved up and handleSFUMigration can be called inside if to avoid duplication

KaustubhKumar05 (approved):

packages/hms-video-store/src/transport/index.ts
raviteja83: it might fail. migration should run only when prev sfuNodeId is present. if not it might run the first time it was set.

eswarclynn (approved):

PR #3149, opened 2024-08-02
fix: mweb tile getting blank when joined from 2 tabs

amar-1995:

-
-

### DIFF
--- a/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
+++ b/packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
@@ -526,7 +526,6 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
         this.nativeTrack?.stop();
         this.nativeTrack = track;
       } else {
-        this.nativeTrack.enabled = false;
         await this.replaceSender(this.nativeTrack, false);
       }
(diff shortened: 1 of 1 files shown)


KaustubhKumar05 (approved):

packages/hms-video-store/src/media/tracks/HMSLocalVideoTrack.ts
```diff
@@ -520,19 +520,29 @@ export class HMSLocalVideoTrack extends HMSVideoTrack {
...
-      this.nativeTrack.enabled = false;
-      HMSLogger.d(this.TAG, 'visibility hidden muting track');
-      this.replaceSenderTrack(this.nativeTrack);
+      if (this.enabled) {
+        const track = await this.replaceTrackWithBlank();
+        await this.replaceSender(track, this.enabled);
+        this.nativeTrack?.stop();
+        this.nativeTrack = track;
+      } else {
+        this.nativeTrack.enabled = false;
```
eswarclynn, line 529: Why is this required? We're in this loop because this.enabled which is essentially this.nativeTrack.enabled is already false?

KaustubhKumar05 (approved):

eswarclynn (approved):

PR #3151, opened 2024-08-02
fix: properly stop all previous tracks during sfu migration

raviteja83:

-
-

### DIFF
--- a/packages/hms-video-store/src/connection/HMSConnection.ts
+++ b/packages/hms-video-store/src/connection/HMSConnection.ts
@@ -28,6 +28,8 @@ export default abstract class HMSConnection {
    *  - [HMSSubscribeConnection] clears this list as soon as we call [addIceCandidate]
    */
   readonly candidates = new Array<RTCIceCandidateInit>();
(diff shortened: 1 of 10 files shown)


packages/hms-video-store/src/connection/HMSConnection.ts
```diff
@@ -28,6 +28,8 @@ export default abstract class HMSConnection {
    *  - [HMSSubscribeConnection] clears this list as soon as we call [addIceCandidate]
    */
   readonly candidates = new Array<RTCIceCandidateInit>();
+  // @ts-ignore
```
KaustubhKumar05, line 31: Why was this needed?

KaustubhKumar05 (approved):

packages/hms-video-store/src/connection/HMSConnection.ts
raviteja83: property is not read here hence needed

eswarclynn (approved):

KaustubhKumar05 (approved):

PR #3161, opened 2024-08-08
fix: clear authToken on leave

raviteja83:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/AuthToken.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/AuthToken.tsx
@@ -2,12 +2,16 @@ import React, { useEffect, useRef, useState } from 'react';
 import { useSessionStorage } from 'react-use';
 import { match } from 'ts-pattern';
 import { v4 as uuid } from 'uuid';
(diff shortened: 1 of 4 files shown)


eswarclynn (approved):

amar-1995 (approved):

packages/roomkit-react/src/Prebuilt/components/AuthToken.tsx
```diff
@@ -20,33 +24,48 @@ import { APP_DATA } from '../common/constants';
...
     }
 
-    if (!savedUserId && !userId) {
-      setSavedUserId(uuid());
+    if (!roomCode) {
+      console.error('room code not provided');
       return;
     }
 
+    let userIdForAuthToken = userId || savedUserId;
```
KaustubhKumar05, line 55: Can be simplified to this?
```suggestion
    let userIdForAuthToken = userId || savedUserId || uuid();
```

raviteja83: it has to be created only once and to be store in storage when we don't have userId

PR #3162, opened 2024-08-08
fix: local peer stats initialisation after migration

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/diagnostics/DiagnosticsStatsCollector.ts
+++ b/packages/hms-video-store/src/diagnostics/DiagnosticsStatsCollector.ts
@@ -35,11 +35,7 @@ export class DiagnosticsStatsCollector {
       localVideoTrackID && this.localVideoTrackStatsList.push(localTrackStats[localVideoTrackID]);
     }
 
-    const subscribeStatsReport = await this.sdk
-      .getWebrtcInternals()
-      ?.getCurrentStats()
-      ?.getSubscribePeerConnection()
-      ?.getStats();
+    const subscribeStatsReport = await this.sdk.getWebrtcInternals()?.getSubscribePeerConnection()?.getStats();
     subscribeStatsReport?.forEach(stat => {
       if (stat.type === 'inbound-rtp') {
         const list = stat.kind === 'audio' ? this.remoteAudioTrackStatsList : this.remoteVideoTrackStatsList;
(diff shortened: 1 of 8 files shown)


packages/roomkit-react/src/Prebuilt/components/StatsForNerds.jsx
```diff
@@ -127,12 +128,12 @@ export const StatsForNerds = ({ onOpenChange }) => {
...
 const useTracksWithLabel = () => {
   const tracksMap = useHMSStore(selectTracksMap);
-  const peersMap = useHMSStore(selectPeersMap);
+  const vanillaStore = useHMSVanillaStore();
   const localPeerID = useHMSStore(selectLocalPeerID);
   const tracksWithLabels = useMemo(
     () =>
       Object.values(tracksMap).reduce((res, track) => {
-        const peerName = peersMap[track.peerId]?.name;
+        const peerName = vanillaStore.getState(selectPeerNameByID(track.peerId));
```
raviteja83, line 136: does it make a difference compared to peermap?

packages/hms-video-store/src/diagnostics/DiagnosticsStatsCollector.ts
```diff
@@ -35,7 +35,11 @@ export class DiagnosticsStatsCollector {
       localVideoTrackID && this.localVideoTrackStatsList.push(localTrackStats[localVideoTrackID]);
     }
 
-    const subscribeStatsReport = await this.sdk.getWebrtcInternals()?.getSubscribePeerConnection()?.getStats();
+    const subscribeStatsReport = await this.sdk
+      .getWebrtcInternals()
+      ?.getCurrentStats()
```
raviteja83, line 40: this chaining is too lengthy too many places/ anyway to reduce?

packages/roomkit-react/src/Prebuilt/components/StatsForNerds.jsx
eswarclynn: the fn in useMemo would be recalculated less number of times?

raviteja83 (approved):

packages/roomkit-react/src/Prebuilt/components/StatsForNerds.jsx
raviteja83: if it is only for name vanilla store js fine, but why does memo have  no deps?

KaustubhKumar05 (approved):

PR #3168, opened 2024-08-09
VB without sidepane

KaustubhKumar05:

- VB is initialised at approutes to not be dependent on the sidepane


yarn.lock
```diff
@@ -7,6 +7,53 @@
   resolved "https://registry.yarnpkg.com/@100mslive/hms-noise-cancellation/-/hms-noise-cancellation-0.0.1.tgz#037c8bdfb6b2d7bf12f9d257422150fe6ca43acb"
   integrity sha512-DGnzcXRDJREWypIjGX70er+f2k/XLLRF41lrXPs1+PtB1imdEkECPPS0Fg4BA0BCWKDNAGTZBHZPrBDgUmr9Lw==
 
+"@100mslive/roomkit-react@0.3.16":
```
raviteja83, line 10: ???? should not be there in lock file I think

PR #3169, opened 2024-08-09
feat: added default VB without opening side panel in preview and meeting room

shivam-kumar-azodha:

- It allows default virtual background to be set automatically without sidepanel to be opened.

### Implementation note, gotchas, related work and Future TODOs (optional)
- It takes 1-2 seconds to download the VB image, which is configurable in 100ms config.


shivam-kumar-azodha:
Mistakenly created PR.

PR #3203, opened 2024-08-27
fix: set autoplay in SDK to avoid play/pause overlay in apple low power mode

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/media/tracks/HMSVideoTrack.ts
+++ b/packages/hms-video-store/src/media/tracks/HMSVideoTrack.ts
@@ -110,5 +110,7 @@ export class HMSVideoTrack extends HMSTrack {
       element.autoplay = true;
     }
     element.playsInline = true;
+    element.muted = true;
+    element.controls = false;
   }
 }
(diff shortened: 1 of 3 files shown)


KaustubhKumar05 (approved):

packages/roomkit-react/src/Video/Video.tsx
```diff
@@ -57,5 +57,5 @@ interface Props {
 
 export const Video: React.FC<Props & StyledProps> = ({ trackId, attach, ...props }) => {
   const { videoRef } = useVideo({ trackId, attach });
-  return <StyledVideo autoPlay muted playsInline controls={false} ref={videoRef} {...props} />;
+  return <StyledVideo muted controls={false} ref={videoRef} {...props} />;
```
raviteja83, line 60: let us move everything then. muted should also be moved into the sdk

eswarclynn: What about controls?

packages/hms-video-store/src/media/tracks/HMSVideoTrack.ts
```diff
@@ -70,15 +71,16 @@ export class HMSVideoTrack extends HMSTrack {
       if (existingTrack?.id === track.id) {
         if (!existingTrack.muted && existingTrack.readyState === 'live') {
           // it's already attached, attaching again would just cause flickering
-          this.reTriggerPlay({ videoElement });
-          return;
+          return this.reTriggerPlay({ videoElement });
```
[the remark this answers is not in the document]
eswarclynn: I'm not sure what case this is, when do we re-attach?

raviteja83 (approved):

PR #3209, opened 2024-09-02
fix: vb code split

KaustubhKumar05:


raviteja83 (approved):

packages/roomkit-react/src/Prebuilt/components/VirtualBackground/VBToggle.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect } from 'react';
-import { HMSVirtualBackgroundTypes } from '@100mslive/hms-virtual-background';
+// eslint-disable-next-line
+import { HMSVirtualBackgroundTypes } from '@100mslive/hms-virtual-background/hmsvbplugin';
```
KaustubhKumar05, line 3: Updating import so this works in older safari versions

amar-1995 (approved):

PR #3236, opened 2024-09-05
fix: select reason while clicking on whole reasons

amar-1995:


packages/roomkit-react/src/Prebuilt/components/EndCallFeedback/FeedbackForm.tsx
```diff
@@ -302,11 +302,11 @@ export const FeedbackForm = ({
                     r: '$1',
                     p: '$6',
                   }}
+                  onClick={() => handleCheckedChange(index)}
```
raviteja83, line 305: we should use label and htmlFor instead. That should handle this.

KaustubhKumar05 (approved):

raviteja83 (approved):

PR #3246, opened 2024-09-08
fix: microphone not released after leave due to race condition

raviteja83:

-
-


packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -271,6 +281,8 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
     await this.pluginsManager.closeContext();
     this.transceiver = undefined;
     this.processedTrack?.stop();
+    this.tracksCreated.forEach(track => track.stop());
+    this.tracksCreated.clear();
```
KaustubhKumar05, line 285: Should we re-check device change scenarios?
cc @amar-1995

amar-1995: Yes we should check that.

eswarclynn (approved):

KaustubhKumar05 (approved):

PR #3262, opened 2024-09-12
fix: added prod event endpoint

amar-1995:


packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -145,7 +145,7 @@ export const HMSPrebuilt = React.forwardRef<HMSPrebuiltRefType, HMSPrebuiltProps
       | undefined;
     const tokenByRoomCodeEndpoint: string = endpointsObj?.tokenByRoomCode || '';
     const initEndpoint: string = endpointsObj?.init || '';
-    const eventEndpoint: string = endpointsObj?.event || '';
+    const eventEndpoint: string | undefined = endpointsObj?.event;
```
raviteja83, line 148: typing here probably not needed. should auto derive

packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -145,7 +145,7 @@ export const HMSPrebuilt = React.forwardRef<HMSPrebuiltRefType, HMSPrebuiltProps
       | undefined;
     const tokenByRoomCodeEndpoint: string = endpointsObj?.tokenByRoomCode || '';
     const initEndpoint: string = endpointsObj?.init || '';
```
raviteja83, line 147: why is an empty string here. can we fix for all?

packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -288,7 +288,7 @@ function AppRoutes({
   authTokenByRoomCodeEndpoint,
   defaultAuthToken,
 }: {
-  authTokenByRoomCodeEndpoint: string;
+  authTokenByRoomCodeEndpoint: string | undefined;
```
raviteja83, line 291: ? should be enough like the defaultAuthToken

packages/roomkit-react/src/Prebuilt/components/AuthToken.tsx
```diff
@@ -25,7 +25,7 @@ import { APP_DATA } from '../common/constants';
  * ui_mode=activespeaker => lands in active speaker mode after joining the room
  */
 const AuthToken = React.memo<{
-  authTokenByRoomCodeEndpoint: string;
+  authTokenByRoomCodeEndpoint: string | undefined;
```
raviteja83, line 28: same here `?` should be enough like below props

raviteja83 (approved):

KaustubhKumar05 (approved):

PR #3276, opened 2024-09-18
build: add nvmrc

grubinsky:

-
-


raviteja83:
we wouldn't want to lock the version. it should and can work with multiple node versions.

PR #3277, opened 2024-09-19
fix: device label categorisation on prebuilt

eswarclynn:

-
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Header/common.jsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Header/common.jsx
@@ -79,7 +79,7 @@ export const AudioActions = () => {
     return null;
   }
   let AudioIcon = <SpeakerIcon />;
-  if (currentSelection && currentSelection.label.toLowerCase().includes('bluetooth')) {
+  if (/airpods|buds|wireless|bluetooth/gi.test(currentSelection.label)) {
     AudioIcon = <BluetoothIcon />;
   } else if (currentSelection && currentSelection.label.toLowerCase().includes('wired')) {
     AudioIcon = <HeadphonesIcon />;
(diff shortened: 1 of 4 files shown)


packages/roomkit-react/src/Prebuilt/components/Header/common.jsx
```diff
@@ -79,7 +79,7 @@ export const AudioActions = () => {
     return null;
   }
   let AudioIcon = <SpeakerIcon />;
-  if (currentSelection && currentSelection.label.toLowerCase().includes('bluetooth')) {
+  if (/airpods|buds|wireless|bluetooth/gi.test(currentSelection.label)) {
```
raviteja83, line 82: should we expose from sdk?

eswarclynn: was thinking of the same, will implement

packages/roomkit-react/src/Prebuilt/components/Header/common.jsx
```diff
@@ -78,12 +79,13 @@ export const AudioActions = () => {
   if (!audioFiltered) {
     return null;
   }
+  const deviceCategory = getAudioDeviceCategory(currentSelection.label);
   let AudioIcon = <SpeakerIcon />;
-  if (currentSelection && currentSelection.label.toLowerCase().includes('bluetooth')) {
+  if (deviceCategory === 'bluetooth') {
```
raviteja83, line 84: should we create enum for the right side values?

eswarclynn: Do we actually need it since the fn returns only certain strings? ` "bluetooth" | "speakerhone" | "wired" | "earpiece" | "speakerphone"`

raviteja83: should be fine for now but there is always a chance for a typo when using it.

raviteja83 (approved):

PR #3291, opened 2024-09-26
fix: role change not happening from broadcaster to hls

amar-1995:


packages/roomkit-react/src/Prebuilt/components/Notifications/Notifications.tsx
```diff
@@ -115,7 +114,6 @@ export function Notifications() {
           ToastManager.addToast({
             title: `You are now a ${notification.data.roleName}`,
           });
-          updateRoomLayoutForRole?.(notification.data.roleName);
```
raviteja83, line 118: why is this not working? is the role not correct here?

packages/roomkit-react/src/Prebuilt/components/Notifications/RoleChangeNotification.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { HMSNotificationTypes, useHMSNotifications } from '@100mslive/react-sdk';
+import { useUpdateRoomLayout } from '../../provider/roomLayoutProvider';
+// @ts-ignore: No implicit Any
+import { ToastManager } from '../Toast/ToastManager';
+// @ts-ignore: No implicit Any
+import { useIsNotificationDisabled } from '../AppData/useUISettings';
+
+export const RoleChangeNotification = () => {
+  const notification = useHMSNotifications(HMSNotificationTypes.ROLE_UPDATED);
+  const isNotificationDisabled = useIsNotificationDisabled();
```
raviteja83, line 11: this is probably not required as it is handled in parent component

packages/roomkit-react/src/Prebuilt/components/Notifications/Notifications.tsx
amar-1995: In some scenarios, we are not getting useIsNotification or the notification is itself is empty. So this will never reach.

packages/roomkit-react/src/Prebuilt/components/Notifications/RoleChangeNotification.tsx
amar-1995: I thought to completely remove it from the parent and handle everything in this file as it will be single point if anything happens when the roleChange notification arrives.

raviteja83: no it should be at the parent as single point, not in each individual component

packages/roomkit-react/src/Prebuilt/components/Notifications/RoleChangeNotification.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const updateRoomLayoutForRole = useUpdateRoomLayout();
+
+  useEffect(() => {
+    if (notification && !isNotificationDisabled) {
```
raviteja83, line 15: this condition is not required. we anyways get this only for local peer.

packages/roomkit-react/src/Prebuilt/components/Notifications/RoleChangeNotification.tsx
```diff
@@ -0,0 +1,26 @@
...
+      ToastManager.addToast({
+        title: `You are now a ${notification.data.roleName}`,
+      });
+    }
+    if (notification && notification.data?.isLocal && notification.data?.roleName) {
```
raviteja83, line 20: isLocal check also should not be needed.

packages/roomkit-react/src/Prebuilt/components/Notifications/RoleChangeNotification.tsx
raviteja83, line 15: why will this work and not the earlier implementation? i don't see a clear difference.

raviteja83: do we get the notification for all peers here?

PR #3296, opened 2024-10-01
feat: upgrade stitches version to 1.3.1-1

hdz-666:

-
-

### DIFF
--- a/packages/roomkit-react/src/Avatar/Avatar.tsx
+++ b/packages/roomkit-react/src/Avatar/Avatar.tsx
@@ -13,7 +13,7 @@ export const StyledAvatar = styled('div', {
   ...flexCenter,
   color: '$colors$on_primary_high',
   fontFamily: '$sans',
-  aspectRatio: 1,
+  aspectRatio: '1 / 1',
   fontWeight: 600,
   fontSize: '$space$9',
   variants: {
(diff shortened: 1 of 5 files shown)


packages/roomkit-react/src/Avatar/Avatar.tsx
```diff
@@ -13,7 +13,7 @@ export const StyledAvatar = styled('div', {
   ...flexCenter,
   color: '$colors$on_primary_high',
   fontFamily: '$sans',
-  aspectRatio: 1,
+  aspectRatio: '1 / 1',
```
KaustubhKumar05, line 16: Is this needed? There are lots of places where we are giving numeric values. Will all of them have to be replaced?

raviteja83: just '1' should be enough. was there a type error @hdz-666 ?

raviteja83: @hdz-666 ??

hdz-666: it was throwing error for 1/1

KaustubhKumar05: It should work with just '1'. Can you update other places where we use numbers?

raviteja83 (approved):

PR #3310, opened 2024-10-08
fix: make role change notification an independent component

ygit:

- fix: WEB-3034 made role change an independent component


amar-1995 (approved):
LGTM

raviteja83 (approved):

PR #3320, opened 2024-10-14
fix: handle terminal error thrown from retrying first error

eswarclynn:

-
-

### DIFF
--- a/packages/hms-video-store/src/transport/RetryScheduler.ts
+++ b/packages/hms-video-store/src/transport/RetryScheduler.ts
@@ -114,17 +114,7 @@ export class RetryScheduler {
       }
     }
 
-    const timeElapsedSinceError = Date.now() - failedAt;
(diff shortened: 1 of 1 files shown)


packages/hms-video-store/src/transport/RetryScheduler.ts
```diff
@@ -158,11 +161,21 @@ export class RetryScheduler {
       taskSucceeded = await this.setTimeoutPromise(task, delay);
     } catch (ex) {
       taskSucceeded = false;
-      HMSLogger.w(
-        this.TAG,
-        `[${TFC[category]}] Un-caught exception ${(ex as HMSException).name} in retry-task, initiating retry`,
-        ex,
-      );
+
+      if ((ex as HMSException).isTerminal) {
```
raviteja83, line 165: `const error = ex as HMSException` before this and reuse everywhere.

raviteja83 (approved):

KaustubhKumar05 (approved):

PR #3321, opened 2024-10-15
fix: tooltip not getting dismissed 

hdz-666:

- event propogation is causing this bubble to happen, stop it after event triggered is bubbled and problem will be solved ;
- 
similar to :  https://github.com/radix-ui/primitives/issues/2248
-

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/DesktopOptions.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/DesktopOptions.tsx
@@ -1,4 +1,4 @@
-import React, { FocusEvent, Fragment, useState } from 'react';
+import React, { Fragment, useState } from 'react';
 import { HMSHLSPlayer } from '@100mslive/hls-player';
 import {
   ConferencingScreen,
@@ -120,12 +120,7 @@ export const DesktopOptions = ({
         onOpenChange={value => updateState(MODALS.MORE_SETTINGS, value)}
         modal={false}
       >
-        <Tooltip
-          title="More options"
-          onFocus={(event: FocusEvent<Element>) => {
-            event.preventDefault();
-          }}
-        >
+        <Tooltip title="More options">
           <Dropdown.Trigger asChild data-testid="more_settings_btn">
             <IconButton>
               <HamburgerMenuIcon />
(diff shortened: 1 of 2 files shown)


packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/DesktopOptions.tsx
```diff
@@ -120,7 +120,12 @@ export const DesktopOptions = ({
         onOpenChange={value => updateState(MODALS.MORE_SETTINGS, value)}
         modal={false}
       >
-        <Tooltip title="More options">
+        <Tooltip
+          title="More options"
+          onFocus={(event: FocusEvent<Element>) => {
+            event.preventDefault();
```
hdz-666, line 126: event propogation is causing this bubble to happen, stop it after event is sent is bubbled and problem will be solved : https://github.com/radix-ui/primitives/issues/2248

raviteja83: should we add it at the tooltip level? we can safely call it on every tooltip right? why do we need to pass a fn where we use tooltip?
did we also check most places where tooltip is used? this was just given as example as it is easy to repro

KaustubhKumar05: I think Chat and any other button that toggles something (sidepane, dropdown) and has a tooltip has this issue

hdz-666: ok wil check this @raviteja83  just thought happening here

KaustubhKumar05: 👀

packages/roomkit-react/src/Tooltip/Tooltip.tsx
```diff
@@ -61,13 +61,12 @@ export const Tooltip: React.FC<
...
   disabled = false,
   delayDuration = 200,
-  ...props
 }) => {
   const [open, setOpen] = useState(false);
 
   return (
     <BaseTooltip.Provider>
-      <TooltipRoot delayDuration={delayDuration} open={open && !disabled} onOpenChange={setOpen} {...props}>
+      <TooltipRoot delayDuration={delayDuration} open={open && !disabled} onOpenChange={setOpen}>
```
raviteja83, line 69: why did we remove the props? are there no other props tooltip takes?

raviteja83 (approved):

KaustubhKumar05 (approved):

PR #3324, opened 2024-10-15
fix: add delay when whiteboard is open

amar-1995:


packages/react-sdk/src/hooks/useWhiteboard.ts
```diff
@@ -34,7 +35,9 @@ export const useWhiteboard = (isMobile = false) => {
     if (open) {
       isOwner && (await actions.interactivityCenter.whiteboard.close());
     } else {
+      setLoading(true);
       await actions.interactivityCenter.whiteboard.open();
+      setTimeout(() => setLoading(false), 500);
```
raviteja83, line 40: can we proper state as discussed in the morning. what if this is not enough? and grpc connection takes more time?

raviteja83, line 40: why is a timeout needed if we are doing it after the open call?

amar-1995: We are maintaining the state of open/close in locally which result into issue. We need someway to know(from backend) that server initiated the connection.

amar-1995: Moved the logic to prebuilt side.

packages/roomkit-react/src/Prebuilt/components/Footer/WhiteboardToggle.tsx
```diff
@@ -25,17 +26,23 @@ export const WhiteboardToggle = () => {
...
             return;
           }
           try {
-            await toggle();
+            if (!open) {
+              setLoading(true);
+              await toggle();
+              setTimeout(() => setLoading(false), 500);
+            } else {
+              await toggle();
```
raviteja83, line 38: should we not show loading while closing? is it instant? or it is fine since we open new connection? for consistency we can use for both

amar-1995: So basically close is instant. We immediately send it close to the backend and which closes grpc session.

raviteja83: for consistency we can still keep it for both right? it would look odd it is disabled only while opening.

KaustubhKumar05: Do we really need to add a delay for closing? This change is in line with screenshare - takes some time to start but stops instantly

raviteja83 (approved):

PR #3333, opened 2024-10-22
Fix: aspect ratio ignored due to type

KaustubhKumar05:
Following the stitches upgrade, aspect ratio cannot be passed as a number

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Preview/PreviewJoin.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Preview/PreviewJoin.tsx
@@ -203,8 +203,6 @@ export const PreviewTile = ({ name, error }: { name: string; error?: boolean }) 
     [calculatedWidth, calculatedHeight],
   );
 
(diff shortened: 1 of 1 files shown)


packages/roomkit-react/src/Prebuilt/components/Preview/PreviewJoin.tsx
```diff
@@ -61,7 +61,7 @@ const useLocalTileAspectRatio = () => {
   } else {
     aspectRatio = isMobile ? 9 / 16 : 16 / 9;
   }
-  return aspectRatio;
+  return aspectRatio.toString();
```
raviteja83, line 64: can't we return in backticks? what does toString do here? give float value?

raviteja83: we can make it string wherr it is passed to css no?

raviteja83 (approved):

hdz-666 (approved):

PR #3355, opened 2024-11-04
fix: remove autohide controls for mweb

KaustubhKumar05:
- For the next release
- Remove tap to toggle header and footer behaviour

## Test scenarios
- Mweb HLS stream - viewer roles (Should remain hidden, same behaviour as prod)
- Mweb conference and broadcaster roles (Should always be visible, no autohide)


ygit (approved):
LGTM

hdz-666 (approved):

PR #3356, opened 2024-11-05
[DO NOT MERGE] feat/WEB-3065 limit theme to container

hdz-666:

-
-


packages/roomkit-react/src/Theme/ThemeProvider.tsx
```diff
@@ -64,13 +66,17 @@ export const HMSThemeProvider: React.FC<React.PropsWithChildren<ThemeProviderPro
     if (!isBrowser) {
       return updatedTheme;
     }
+    let element = document.querySelector(container);
```
KaustubhKumar05, line 69: This throws an error if container is an empty string. Let's add a check

PR #3357, opened 2024-11-07
fix: corrected typo in Flex component

ygit:

- fix: corrected typo in Flex component


raviteja83 (approved):

KaustubhKumar05:
We should mention this in the release changelog since it could break UIs

PR #3358, opened 2024-11-07
fix: clean up VB app data state on reset

KaustubhKumar05:


packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -131,6 +132,7 @@ export const HMSPrebuilt = React.forwardRef<HMSPrebuiltRefType, HMSPrebuiltProps
       // leave room when component unmounts
       return () => {
         VBHandler.reset();
```
raviteja83, line 134: why can't this handled inside the reset fn?

KaustubhKumar05: VBHandler is a class and I need to access hmsActions

KaustubhKumar05: Can try passing a callback resetting appData to VBHander and it can call it on reset. Does that work?

raviteja83 (approved):

packages/roomkit-react/src/Prebuilt/AppStateContext.tsx
```diff
@@ -79,6 +80,7 @@ export const useAppStateManager = () => {
               .otherwise(() => PrebuiltStates.MEETING);
           });
           VBHandler.reset();
+          hmsActions.setAppData(APP_DATA.background, HMSVirtualBackgroundTypes.NONE);
```
hdz-666, line 83: Are we sure this is not leading to multiple renders ?

KaustubhKumar05: Yes, since this is not a dependency and the peer gets redirected to the leave screen after this. The sidepane is unmounted

packages/roomkit-react/src/Prebuilt/App.tsx
```diff
@@ -130,7 +129,6 @@ export const HMSPrebuilt = React.forwardRef<HMSPrebuiltRefType, HMSPrebuiltProps
     useEffect(() => {
       // leave room when component unmounts
       return () => {
-        VBHandler.reset();
```
hdz-666, line 133: why did we remove this ? should it be dismounted at end ?

KaustubhKumar05: This is redundant. Had initially added it here to clean up the state on rejoin. Since we moved to the appstate router just calling it there is sufficient. Forgot to remove it from here then

ygit (approved):

PR #3361, opened 2024-11-08
fix: add validations in getDebugInfo

KaustubhKumar05:


ygit (dismissed):
LGTM

packages/roomkit-react/src/Prebuilt/components/MoreSettings/SplitComponents/MwebOptions.tsx
```diff
@@ -327,7 +327,9 @@ export const MwebOptions = ({
         </Sheet.Content>
       </Sheet.Root>
       <SettingsModal open={openSettingsSheet} onOpenChange={setOpenSettingsSheet} screenType={screenType} />
-      <StatsForNerds open={openStatsForNerdsSheet} onOpenChange={setOpenStatsForNerdsSheet} />
+      {openStatsForNerdsSheet && (
+        <StatsForNerds open={openStatsForNerdsSheet} onOpenChange={setOpenStatsForNerdsSheet} />
```
raviteja83, line 331: can't we do the check inside the component

KaustubhKumar05: Would be better to keep it outside in this case imo, like how we're doing just below this. It will affect the readability inside the component even more

raviteja83 (dismissed):

packages/hms-video-store/src/sdk/index.ts
```diff
@@ -281,7 +281,11 @@ export class HMSSdk implements HMSInterface {
     return this.transport?.getWebrtcInternals();
   }
 
-  getDebugInfo(): DebugInfo {
+  getDebugInfo(): DebugInfo | undefined {
+    if (!this.transport) {
+      HMSLogger.d(this.TAG, `Transport is not defined`);
```
raviteja83, line 286: better if we throw an error error,

hdz-666 (approved):

PR #3371, opened 2024-11-14
feat: internal method to check if recording should continue

KaustubhKumar05:
# Description

Beam currently continues recording even when it is not able to subscribe to any peers. Recording gets marked as successful despite being blank. Adding this method so beam can identify this and terminate after a timeout if the check fails continuously. 

This should not cause regular flows to fail (RTMP IN - no peers needed to join, rooms where js bridge is disabled - won't be able to execute this)

https://app.devrev.ai/100ms/works/ISS-30657

Usage:

image

---

### Pre-launch Checklist

### Merging:
- Squash merge to dev


hdz-666 (approved):

ygit (approved):
LGTM

PR #3415, opened 2024-12-11
doximity rtc peer connection keyword modified to public from readonly

hdz-666:
# Description
image
---

Exposed nativeConnection class for getting rtc peer connection 

__hms.sdk.transport.subscribeConnection
__hms.sdk.transport.publishConnection

### Pre-launch Checklist

- [x] I updated/added relevant documentation.

### Merging:
- Squash merge to dev

### DIFF
--- a/packages/hms-video-store/src/connection/HMSConnection.ts
+++ b/packages/hms-video-store/src/connection/HMSConnection.ts
@@ -11,12 +11,27 @@ import { enableOpusDtx, fixMsid } from '../utils/session-description';
 
 const TAG = '[HMSConnection]';
 
+/**
+ * Abstract base class that handles WebRTC peer connections for 100ms SDK
+ * Provides common functionality for both publishing (sending) and subscribing (receiving) connections
+ */
 export default abstract class HMSConnection {
+  /**
+   * Indicates whether this connection is for publishing or subscribing media
+   */
   readonly role: HMSConnectionRole;
+
   protected readonly signal: JsonRpcSignal;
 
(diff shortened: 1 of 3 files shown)


KaustubhKumar05:
@raviteja83 are these changes sufficient?

packages/hms-video-store/src/connection/HMSConnection.ts
```diff
@@ -11,12 +11,27 @@ import { enableOpusDtx, fixMsid } from '../utils/session-description';
...
 
   protected abstract readonly observer: IConnectionObserver;
-  abstract readonly nativeConnection: RTCPeerConnection;
+
+  /**
+   * The native WebRTC peer connection object
+   * @abstract
+   * @type {RTCPeerConnection}
+   */
+  abstract nativeConnection: RTCPeerConnection;
```
raviteja83, line 33: this can be readonly so that it shouldn't be allowed to be modified. They just need a read access which we don't expose directly.

raviteja83:
```
 private getNativePublishPeerConnection() {
    return this.transport.getPublishConnection()?.nativeConnection;
  }
```
Just adding two methods -  one for publish and other for subscribe in sdk/index and HMSSDKActions should be enough. we can even combine this to getNativePeerConnections and return and object with publish and subscribe

raviteja83: we should add validations whether join has happened or not and throwing an error suggesting to call post joiin.

PR #3451, opened 2025-01-30
fix: improve button hover and focus colors

### LINKED ISSUE #3450: [Bug]: Red danger buttons should have visible text on hover
### What happened?

When red danger buttons are hovered right now, the text is not visible:

Image
Image
Image
Image
Image
Image
Image

Also, when they are focused, there is currently no focus border

### How can we reproduce the bug?

- hover over any danger button

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

aldenquimby:
# Description

Fixes #3450 

When danger buttons are hovered, the text is now visible:
Image
Image
Image
Image
Image

The focus indicator also works now:
hover-danger

And the focus change applied to all solid (non-outlined) buttons. Focus was previously invisible and is now shown:
hover

---

-

### Pre-launch Checklist
- [x] I updated/added relevant documentation.

### Merging:
- Squash merge to dev


packages/roomkit-react/src/Button/Button.tsx
```diff
@@ -70,7 +74,7 @@ const getButtonVariants = (
       border: `1px solid ${active}`,
     },
     '&:not([disabled]):focus-visible': {
-      boxShadow: `0 0 0 3px ${base}`,
+      boxShadow: `0 0 0 3px $colors${hover}`,
```
aldenquimby, line 77: 1. `${base}` was rendering an invalid CSS var here. Add `$colors` fixed that
2. `${base}` is the background color, so it was blending in. using `${hover}` instead is clearer

raviteja83 (approved):

PR #3459, opened 2025-02-03
feat(prebuilt): allow custom footer in participant side pane

### LINKED ISSUE #3446: [Feature]: Prebuilt should allow custom react component in participants side pane
### What's the feature?

My company would like to add some custom components to the Participants side pane, without needing to maintain a separate fork of HMSPrebuilt

Example use cases:
- adding a "copy invite link" component, similar to what Zoom has, making it easier to share the current meet URL
    - I realize you could copy the current browser URL, but our users are not very tech savvy
- adding a "visibility setting" control to toggle between Public and Private meetings
    - The auth for this would be handled within my app, outside of 100ms

aldenquimby:
# Description
Fixes #3446

Allows prebuilt consumers to render a custom react component at the bottom of the participant list side pane:
image

- I could not figure out where `types-prebuilt` is coming from
- I suspect using `ReactNode` as the type may be problematic b/c that library doesn't have `react` as a dependency right now

### Pre-launch Checklist

### Merging:
- Squash merge to dev

### DIFF
--- a/packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
+++ b/packages/roomkit-react/src/Prebuilt/components/Footer/ParticipantList.tsx
@@ -46,9 +46,11 @@ import { SIDE_PANE_OPTIONS } from '../../common/constants';
 export const ParticipantList = ({
   offStageRoles = [],
   onActive,
+  footer,
 }: {
   offStageRoles: HMSRoleName[];
   onActive: (role: string) => void;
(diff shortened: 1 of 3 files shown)


raviteja83:
we would recommend to do this on a fork. we will add it here if there are few more requests for this.

aldenquimby:
Sounds good. Appreciate you merging all the other PRs @raviteja83 !

PR #3504, opened 2025-05-14
feat: stop reason for hls stop

KaustubhKumar05:
# Description

Added the option to send stop_reason while stopping HLS. Pending changes in biz and beam 

### Pre-launch Checklist

- [x] WS message being sent should contain the stop reason

### DIFF
--- a/packages/hms-video-store/src/interfaces/hls-config.ts
+++ b/packages/hms-video-store/src/interfaces/hls-config.ts
@@ -19,6 +19,10 @@ export interface HLSConfig {
   };
 }
 
+export interface StopHLSConfig extends HLSConfig {
+  stop_reason?: string;
+}
+
 export interface HLSMeetingURLVariant {
   /**
    * This meeting URL is opened in a headless chrome instance for generating the HLS feed.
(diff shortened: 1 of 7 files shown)


packages/hms-video-store/src/interfaces/hls-config.ts
```diff
@@ -19,6 +19,10 @@ export interface HLSConfig {
   };
 }
 
+export interface StopHLSConfig extends HLSConfig {
+  stop_reason?: string;
```
raviteja83, line 23: can be just reason given the config is named StopHLS already

packages/roomkit-react/src/Prebuilt/components/Leave/LeaveRoom.tsx
```diff
@@ -42,11 +42,11 @@ export const LeaveRoom = ({
   const isMobileHLSStream = useMobileHLSStream();
   const isLandscapeHLSStream = useLandscapeHLSStream();
 
-  const stopStream = async () => {
+  const stopStream = async (stop_reason = '') => {
```
raviteja83, line 45: take it as an object so it will be clear when calling it what field is sent

PR #3546, opened 2025-10-28
fix: audio output selecting wrong device, remove unavailable devices that are stored

raviteja83:
# Description

- Correct audio output device is not getting selected in safari. 
- Restricting setSinkId only for chromium based browsers as it causing robotic voice and other issues.
- Removing stored device if they no longer exist which is causing an error everytime because we are using exact now instead of ideal as chrome 138 stopped supporting ideal.

_List which issues are fixed by this PR. You must list at least one issue._

---

-

### Pre-launch Checklist

- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] All existing and new tests are passing.

### Merging:
- Squash merge to dev

### DIFF
--- a/packages/hms-video-store/package.json
+++ b/packages/hms-video-store/package.json
@@ -51,7 +51,7 @@
     "lodash.isequal": "^4.5.0",
     "reselect": "4.0.0",
     "sdp-transform": "^2.14.1",
-    "ua-parser-js": "^1.0.1",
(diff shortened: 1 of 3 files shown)


raviteja83:
@aldenquimby most likely, yes. were you able to upgrade, did it fix the issue? is there a repro you can give on the version you are on. we tried to repro earlier but were not able to consistently repro.

PR #3570, opened 2025-12-18
feat: capture CPU pressure state per sample

raviteja83:
Moves CPU state capture from toAnalytics() (every ~5 mins) to collateSample() (every 30s) for more granular CPU pressure tracking in publish stats.

# Description

_Replace this paragraph with a description of what this PR is changing or adding, and why. Consider including before/after screenshots._

_List which issues are fixed by this PR. You must list at least one issue._

---

-

### Pre-launch Checklist

- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] All existing and new tests are passing.

### Merging:
- Squash merge to dev


raviteja83:
> I see we are using `toAnalytics` to frame the actual event, are we keeping this cpu pressure field through this method?

The samples [], each entry has the cpu state. so it should be there.

PR #3664, opened 2026-04-30
fix(transport): correct role check in handleIceConnectionFailure

raviteja83:
The conditional in `handleIceConnectionFailure` used `HMSConnectionRole.Publish` (numeric `0`, falsy) as a boolean operand instead of `role === HMSConnectionRole.Publish`, so the in-progress guard always inspected `SubscribeIceConnectionFailed`. When a subscribe-side passive recovery is pending, an incoming publish-side ICE failure was silently suppressed — blocking the client-initiated publish iceRestart that is the only recovery path for that side.

### DIFF
--- a/packages/hms-video-store/src/test/unit/transport/handleIceConnectionFailure.test.ts
+++ b/packages/hms-video-store/src/test/unit/transport/handleIceConnectionFailure.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Bug 1 — handleIceConnectionFailure ternary uses HMSConnectionRole.Publish
+ * (numeric 0, falsy) as a conditional, so the in-progress check always
+ * inspects SubscribeIceConnectionFailed regardless of the role argument.
+ *
+ * Real-world consequence:
+ *   - Subscribe-side ICE failure is recovered by a passive wait (server
+ *     re-sends an offer; client just answers).
+ *   - Publish-side ICE failure must be recovered by the client itself
+ *     issuing an iceRestart renegotiation.
+ *   - When BOTH sides fail at the same time (typical of network blips),
+ *     the bug suppresses the publish restart because a subscribe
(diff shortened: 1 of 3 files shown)


AnshulMalik (approved):

raviteja83:
## QA validation

### Static — minified bundle diff

`HMSConnectionRole.Publish === 0`, so the broken ternary `0 ? 3 : 4` constant-folds to a literal `4`:

| | guard call in deployed JS |
|---|---|
| `main` (prod) | `isTaskInProgress(4)` |
| this PR (Vercel preview) | `isTaskInProgress(t===0?3:4)` |

`3` = `PublishIceConnectionFailed`, `4` = `SubscribeIceConnectionFailed`. On `main` the guard is hard-wired to inspect Subscribe regardless of role.

### Dynamic — NLC 100% loss for ~30s

The bug is order-sensitive: it only suppresses Publish recovery when Subscribe disconnects first and Publish disconnects while Subscribe's retry is queued. Real runs against prod and the Vercel preview, both landing Sub-first ordering:

| Time after sub disconnect | `main` (prod) | this PR (Vercel) |
|---|---|---|
| ~5s (Sub `ICE_DISCONNECTION_TIMEOUT` fires) | `retry=[Subscribe]` | `retry=[Subscribe]` |
| ~6s (Pub `ICE_DISCONNECTION_TIMEOUT` fires) | `retry=[Subscribe, SignalDisconnect]` — **Publish absent** | `retry=[Subscribe, Publish]` — `have-local-offer` (iceRestart fires) |

On `main`, Publish never enters the retry queue across the full ~18s recovery window. With the fix, Publish is scheduled ~1s after Subscribe, as the design intends.

A separate run that happened to land **Pub-first** ordering showed no divergence between the two builds — confirming the bug manifests only on Subscribe→Publish failure ordering.

PR #3666, opened 2026-04-30
fix(transport): bail retrySignalDisconnectTask if joinParameters cleared

raviteja83:
\`leave()\` clears \`joinParameters\` synchronously, but the retry scheduler can fire \`retrySignalDisconnectTask\` asynchronously. If the timings cross, the task body dereferences \`this.joinParameters!.authToken\` on \`undefined\`, throwing a \`TypeError\` that propagates to the wrapping retry callback and silently resolves dependents to \`undefined\`. Early-return when \`joinParameters\` is missing or state is \`Leaving\` — the retry scheduler is already being reset by \`leave()\`.

### DIFF
--- a/packages/hms-video-store/src/test/helpers/makeTransport.ts
+++ b/packages/hms-video-store/src/test/helpers/makeTransport.ts
@@ -1,0 +1,68 @@
+/**
+ * Minimal HMSTransport instantiator for unit tests.
+ *
+ * Builds a real `HMSTransport` with the smallest mocks that get past the
+ * constructor and its initial subscriptions. Tests can then call
+ * private/arrow-field methods directly on the instance:
+ *
+ *   const t = makeTransport();
+ *   await (t as any).retrySignalDisconnectTask();
+ *
+ * For tests that exercise specific code paths, override fields after
+ * construction:
+ *
+ *   (t as any).joinParameters = undefined;
+ *   (t as any).signal = mySignalMock;
+ *
+ * This is deliberately not a "fake biz" or full integration harness —
+ * it's the smallest scaffolding that lets us exercise the real
+ * HMSTransport methods with controlled inputs.
+ */
+
+import { AnalyticsEventsService } from '../../analytics/AnalyticsEventsService';
+import { AnalyticsTimer } from '../../analytics/AnalyticsTimer';
+import { PluginUsageTracker } from '../../common/PluginUsageTracker';
+import { DeviceManager } from '../../device-manager';
+import { EventBus } from '../../events/EventBus';
+import { Store } from '../../sdk/store';
+import HMSTransport from '../../transport';
+import ITransportObserver from '../../transport/ITransportObserver';
+import { TransportState } from '../../transport/models/TransportState';
+
+export const makeFakeObserver = (): ITransportObserver => ({
+  onNotification: jest.fn(),
+  onConnected: jest.fn(),
+  onTrackAdd: jest.fn(),
+  onTrackRemove: jest.fn(),
+  onFailure: jest.fn(),
+  onStateChange: jest.fn(async () => {}),
+});
+
+export interface MakeTransportOptions {
+  observer?: ITransportObserver;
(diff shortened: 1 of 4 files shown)


raviteja83:
## Reproducing this on prod (no debug build needed)

The race is reproducible from any browser DevTools console on a 100ms preview/meeting URL — `window.__hms` is already exposed on prod, so we can reach `transport.retrySignalDisconnectTask` directly to simulate the queued-task-body-firing-during-leave race deterministically.

### Steps

1. Open any 100ms preview URL in Chrome with DevTools open.
2. Click **Join Now**, wait until you're actually in the room.
3. Paste this into the console and hit Enter:

```js
(async () => {
  const t = window.__hms?.sdk?.transport;
  const a = window.__hms?.actions;
  if (!t || !a) { console.error('window.__hms.sdk.transport / actions not found — are you on a 100ms preview/meeting page?'); return; }
  if (t.state !== 'Joined') { console.error(`transport.state is "${t.state}", expected "Joined"`); return; }

  const before = { joinParametersSet: t.joinParameters !== undefined, state: t.state, signalConnected: t.signal?.isConnected };

  // 1. Close the signal WS — what a network blip does
  try { t.signal.close(); } catch {}
  // 2. Yield one microtask — lets the retry scheduler queue retrySignalDisconnectTask
  await Promise.resolve();
  // 3. Begin a leave — synchronously clears joinParameters and resets the scheduler
  const leavePromise = a.leave();
  // 4. Simulate the queued retry-task body firing inside the leave critical section
  let raceError = null;
  try { await t.retrySignalDisconnectTask(); }
  catch (err) { raceError = { name: err?.name, message: err?.message, stackTop: (err?.stack || '').split('\n').slice(0, 3).join(' | ') }; }
  try { await leavePromise; } catch {}
  await new Promise(r => setTimeout(r, 100));

  const after = { joinParametersSet: t.joinParameters !== undefined, state: t.state };
  const reproduced = raceError?.name === 'TypeError' && /authToken/.test(raceError?.message || '');
  console.log('%c[Bug 4 race QA] reproduced=' + reproduced, 'font-weight:bold;color:' + (reproduced ? '#d44' : '#494'), { before, after, raceError });
})();
```

### Expected output

**On prod main (broken):**

```
[Bug 4 race QA] reproduced=true
{
  before: { joinParametersSet: true, state: 'Joined', signalConnected: true },
  after:  { joinParametersSet: false, state: 'Disconnected' },
  raceError: {
    name: 'TypeError',
    message: "Cannot read properties of undefined (reading 'authToken')",
    stackTop: "TypeError: Cannot read properties of undefined (reading 'authToken') |     at Y.<anonymous> (.../main.*.js:94:199454) |     at Generator.next"
  }
}
```

The minified `Y.<anonymous>` is `retrySignalDisconnectTask`; the `authToken` access is `transport/index.ts:1283` (`this.joinParameters!.authToken`).

**On a build with this PR's guard:**

```
[Bug 4 race QA] reproduced=false
{
  before: { joinParametersSet: true, state: 'Joined', signalConnected: true },
  after:  { joinParametersSet: false, state: 'Disconnected' },
  raceError: null
}
```

The new early-return (`if (!this.joinParameters || this.state === TransportState.Leaving) return false`) short-circuits the task body before it ever dereferences `joinParameters`.

### Verifying locally without redeploying

If you want to confirm the fix works without rebuilding/redeploying, paste this **before** running the race snippet on the broken page — it patches the same guard onto the live transport instance:

```js
(() => {
  const t = window.__hms.sdk.transport;
  const original = t.retrySignalDisconnectTask.bind(t);
  t.retrySignalDisconnectTask = async function () {
    if (!t.joinParameters || t.state === 'Leaving') return false;
    return original();
  };
  console.log('Guard patched in place — re-run the race snippet to see reproduced=false');
})();
```

I ran this end-to-end on prod (commit 52bc8ca49 deployed) — TypeError on the unmodified page, no error after the in-place guard, identical inputs both times.

cyril-k-031225 (approved):

PR #3681, opened 2026-06-10
LIV-487: show incoming chat messages as a transient bubble in PIP

Shantanu-TAM:
## What

Surfaces the latest **incoming** chat message as a transient, bottom-anchored bubble on the Picture-in-Picture canvas, so users don't miss chat while the tab is backgrounded. The bubble shows the sender name and a single, ellipsised line of text, and auto-dismisses after a 4s TTL.

LIV-487.

## Files at a glance

All changes are confined to `packages/roomkit-react/src/Prebuilt/components/PIP/`.

| File | Role | Key pieces |
|---|---|---|
| `PIPComponent.tsx` | Input / wiring (React) | `ActivatedPIP` effect subscribes to the HMS store's chat messages |
| `pipMessageUtils.ts` | Pure logic | `pickIncomingMessage()` — batch-aware selection of newest incoming text msg |
| `PIPManager.ts` | State + loop (singleton) | `setLatestMessage` / `getActiveMessage` (4s TTL) + the render loop |
| `pipUtils.js` | Rendering (canvas) | `drawVideoElementsOnCanvas` → `drawMessageBubble` (+ `traceRoundedRect`, `truncateToWidth`) |
| `pip.test.js` | Tests | TTL gating + `pickIncomingMessage` cases |

Notes: `pipUtils.js` bumps the canvas 480×320 → 480×360 to give the bubble room, with sizes tuned generously because the browser upscales the PIP window beyond the canvas resolution. `PIPComponent.tsx` seeds with the latest existing message id so only post-PIP messages trigger a bubble.

## Fixes a batching bug

The store appends chat messages in **batches** (the `newMessage` action is debounced ~150ms), so a single subscription fire can carry several new messages. An initial implementation inspected only the tail message, which dropped incoming messages whenever a batch ended with the local user's own or a non-text message — the common multi-party chat case. The selection logic now scans every message appended since the last shown id and surfaces the newest incoming text message, always advancing past the scanned batch. Covered by a regression test.

## Browser support

PIP (and therefore the bubble) is effectively **Chromium-desktop-only**, inheriting the pre-existing `PictureInPicture.isSupported()` gate:

| Browser | PIP option | Bubble | Reason |
|---|---|---|---|
| Chrome / Chromium desktop | shows | ✅ | standard `requestPictureInPicture()` + canvas capture |
| Firefox | hidden | ❌ | no `document.pictureInPictureEnabled` |
| Safari (macOS) / iOS | hidden by design | ❌ | non-standard WebKit PIP; canvas-capture source unsupported |

## Testing

- `yarn workspace @100mslive/roomkit-react types` passes.
- Unit tests added for TTL gating and `pickIncomingMessage` (burst/own-tail regression, filtering rules).
- Manually verified in Chrome: bubble appears bottom-anchored, truncates long text, ignores own messages, clears after ~4s.

### Follow-ups (out of scope)
- No automated coverage yet for the canvas-drawing path (`roomkit-react` has no jest setup; only pure helpers are unit-tested here).
- Pre-existing (not introduced here): `PIPManager.listenToStateChange` has no removal path; the `ActivatedPIP` effect never detaches the state-change listener. Worth a separate cleanup.


packages/roomkit-react/src/Prebuilt/components/PIP/PIPComponent.tsx
```diff
@@ -66,6 +74,23 @@ export const ActivatedPIP = () => {
...
+      // the PIP canvas. Seed with the latest existing message id so only messages
+      // that arrive after PIP starts trigger a bubble.
+      let lastShownMessageId = store.getState(selectHMSMessages)?.slice(-1)[0]?.id;
+      const unsubscribeMessages = store.subscribe(messages => {
+        const result = pickIncomingMessage(messages, lastShownMessageId, store.getState(selectLocalPeerID));
+        lastShownMessageId = result.lastShownMessageId;
+        if (result.message) {
+          PictureInPicture.setLatestMessage(result.message);
+        }
+      }, selectHMSMessages);
```
raviteja83, line 92: we can simply listen to HMSNotificationTypes.NEW_MESSAGE and it would give the latest message. all this logic of pickmessage probably not needed

Shantanu-TAM: There was a bug here, fixed it- check PR desc's this part-

> Fixes a batching bug
> The store appends chat messages in batches (the newMessage action is debounced ~150ms), so a single subscription fire can carry several new messages. An initial implementation inspected only the tail message, which dropped incoming messages whenever a batch ended with the local user's own or a non-text message — the common multi-party chat case. The selection logic now scans every message appended since the last shown id and surfaces the newest incoming text message, always advancing past the scanned batch. Covered by a regression test.
>

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
```diff
@@ -212,7 +246,7 @@ class PipManager {
         return;
       }
       if (this.state === PIPStates.started) {
-        drawVideoElementsOnCanvas(this.videoElements, this.canvas);
+        drawVideoElementsOnCanvas(this.videoElements, this.canvas, this.getActiveMessage());
```
raviteja83: **Critical:** `renderLoop` reschedules only *after* a successful draw and has no try/catch. The new bubble path adds unguarded canvas ops (`measureText`/`arcTo`/`fillText`, plus an unchecked `canvas.getContext('2d')`). A throw here permanently freezes the PIP video on its last frame with no log and no recovery. Wrap this draw in try/catch and reschedule in a `finally`, and/or isolate `drawMessageBubble` + null-check `ctx`.

packages/roomkit-react/src/Prebuilt/components/PIP/pipMessageUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function pickIncomingMessage(
+  messages: HMSMessage[] | undefined,
+  lastShownMessageId: string | undefined,
+  localPeerId: string | undefined,
+): PickIncomingMessageResult {
+  if (!messages?.length) {
+    return { message: null, lastShownMessageId };
+  }
+  const lastShownIndex = messages.findIndex(message => message.id === lastShownMessageId);
```
raviteja83, line 41: When `lastShownMessageId` isn't in the list, `findIndex` returns `-1` and `slice(-1 + 1)` = `slice(0)` rescans the whole history — resurfacing a stale message as a "new" bubble. Reachable if the store prunes the seeded id in a long session, or on a `pinnedTrack` re-subscribe. Handle `-1` explicitly (advance the cursor, show nothing) and add a test for it.

raviteja83 (commented):
## Automated review — PIP incoming-chat bubble

Solid feature — the batch-safe message picker and the TTL logic are correct and well-tested. One critical robustness gap and a few important items (inline comments on the specifics).

### 🔴 Critical
- **Render loop has no error isolation and reschedules only on success.** The new bubble path adds unguarded canvas ops (`measureText`/`arcTo`/`fillText`, unchecked `getContext`); a draw throw permanently freezes PIP video with no log. Wrap the draw in try/catch (reschedule in `finally`) and/or isolate `drawMessageBubble` + null-check `ctx`. A decorative bubble shouldn't be able to kill video rendering.

### 🟡 Important
- `pickIncomingMessage` rescans all history when `lastShownMessageId` isn't found (`findIndex` -1 → `slice(0)`) — can resurface a stale bubble.
- The new messages subscription has no `.catch`, unlike the sibling tracks subscription in the same effect.
- `any` from `HMSMessage.message` flows into the typed `text` field (the `typeof` guard doesn't narrow at the return site).
- **Tests don't run in CI:** `roomkit-react` has no `test` script / jest config, so `lerna run test` skips this file — the new tests give false confidence. Worth wiring up.

### 🟢 Suggestions
- Collapse duplicate `PIPChatMessage` / `PIPIncomingMessage` into one shared type; import the shared `PIP_CANVAS_*` constants instead of redeclaring `DEFAULT_CANVAS_*`; reword "debounce" → "batch" (it's a leading-edge throttle via `ActionBatcher`); extract + unit-test `truncateToWidth` (consider a binary-search shrink).

### Strengths
Reverse-scan correctly fixes the own-message-tail batch bug (with a regression test); TTL is centralized in `getActiveMessage` and tested at the exact boundaries; normalization is pushed to the boundary producer; `traceRoundedRect` deliberately avoids the non-universal `ctx.roundRect()`.

_Generated with Claude Code (multi-agent review)._

packages/roomkit-react/src/Prebuilt/components/PIP/PIPComponent.tsx
```diff
@@ -66,6 +74,23 @@ export const ActivatedPIP = () => {
           console.error('error in updating pip', err);
         });
       }, selectTracksMap);
+
+      // Show incoming chat messages (from other peers) as a transient bubble on
+      // the PIP canvas. Seed with the latest existing message id so only messages
+      // that arrive after PIP starts trigger a bubble.
+      let lastShownMessageId: string | undefined = store.getState(selectHMSMessages)?.slice(-1)[0]?.id;
+      const unsubscribeMessages = store.subscribe(messages => {
```
raviteja83, line 82: Unlike the tracks subscription just above (which does `.catch(err => console.error('error in updating pip', err))`), this callback has no error handling — a malformed message shape would throw silently into the store's notify path. Mirror the sibling's guard with a try/catch + log.

packages/roomkit-react/src/Prebuilt/components/PIP/pipMessageUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+    .find(
+      message =>
+        message.sender &&
+        message.sender !== localPeerId &&
+        typeof message.message === 'string' &&
+        message.message.trim().length > 0,
+    );
+  return {
+    message: latestIncoming
+      ? { senderName: latestIncoming.senderName || 'Anonymous', text: latestIncoming.message }
```
raviteja83: `HMSMessage.message` is typed `any`, and the `typeof === 'string'` guard inside `.find()` doesn't narrow `latestIncoming.message` here, so `any` flows into the `string` `text` field (and `senderName` is passed through untyped). Capture the validated string in a typed local / coerce `senderName` — satisfies the no-`any` rule and protects the downstream `measureText`.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
```diff
@@ -6,8 +6,17 @@ import { isIOS, isMacOS, isSafari } from '../../common/constants';
 const MAX_NUMBER_OF_TILES_IN_PIP = 4;
 const DEFAULT_FPS = 30;
 const DEFAULT_CANVAS_WIDTH = 480;
-const DEFAULT_CANVAS_HEIGHT = 320;
+// Keep in sync with PIP_CANVAS_HEIGHT in pipUtils. A bit taller than the video-only
+// size so the incoming-chat bubble overlays the bottom band without hiding much video.
+const DEFAULT_CANVAS_HEIGHT = 360;
```
raviteja83: `DEFAULT_CANVAS_HEIGHT`/`WIDTH` duplicate `PIP_CANVAS_HEIGHT`/`WIDTH` in `pipUtils` (the "keep in sync" comment only covers height; width drifts too). `PIPManager` already imports from `pipUtils` — import the constants instead of redeclaring to remove the hazard.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
```diff
@@ -6,8 +6,17 @@ import { isIOS, isMacOS, isSafari } from '../../common/constants';
...
 const LEAVE_EVENT_NAME = 'leavepictureinpicture';
+// how long an incoming chat bubble stays on the PIP canvas before it is removed
+const MESSAGE_TTL_MS = 4000;
+
+export interface PIPChatMessage {
```
raviteja83: `PIPChatMessage` is byte-identical to `PIPIncomingMessage` in `pipMessageUtils` and they flow producer→consumer; they only compile together via structural typing. Collapse into one shared type so a future field addition can't silently diverge.

packages/roomkit-react/src/Prebuilt/components/PIP/pipUtils.js
```diff
@@ -181,3 +207,102 @@ function getRenderDimensions(video, width, height) {
...
+  ctx.arcTo(x + width, y + height, x, y + height, r);
+  ctx.arcTo(x, y + height, x, y, r);
+  ctx.arcTo(x, y, x + width, y, r);
+  ctx.closePath();
+}
+
+/**
+ * Truncate text with an ellipsis so it fits within maxWidth for the current ctx font.
+ */
+function truncateToWidth(ctx, text, maxWidth) {
```
raviteja83: `truncateToWidth` is pure and unit-testable under jest-canvas-mock but untested; it also runs `measureText` per character every frame for the TTL window — a binary-search shrink would bound it.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPComponent.tsx
Shantanu-TAM: Deferring this one (not done in `e2a750c`). `HMSNotificationTypes.NEW_MESSAGE` would fire per message and avoid the batch-tail issue, but switching the subscription model is a larger change than the rest of this round, and the current `pickIncomingMessage` path is already batch-correct and unit-tested (incl. the own-message-tail regression). Tracking the `NEW_MESSAGE` simplification as a follow-up rather than reworking it here — happy to do it in a separate PR if you'd prefer that over the selector approach.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
Shantanu-TAM: Fixed in `e2a750c`. The draw is now wrapped in try/catch and the loop reschedules in a `finally`, so a throw logs and the next frame retries instead of freezing PIP. Also null-checked `canvas.getContext('2d')` in `drawVideoElementsOnCanvas` (early-returns if absent).

packages/roomkit-react/src/Prebuilt/components/PIP/pipMessageUtils.ts
Shantanu-TAM: Fixed in `e2a750c`. When the cursor id isn't found (undefined seed or pruned in a long session), it now advances to the tail and shows nothing instead of `slice(0)`-rescanning history. Added a regression test (`when lastShownMessageId is pruned (not found), does not rescan history`).

packages/roomkit-react/src/Prebuilt/components/PIP/PIPComponent.tsx
Shantanu-TAM: Fixed in `e2a750c`. Wrapped the messages-subscription callback in try/catch + `console.error('error surfacing pip chat message', err)`, mirroring the sibling tracks subscription.

packages/roomkit-react/src/Prebuilt/components/PIP/pipMessageUtils.ts
Shantanu-TAM: Fixed in `e2a750c`. The validated string is now captured in a typed local inside the loop (narrowed by the `typeof === 'string'` guard) before it's returned as `text`, and `senderName` is coalesced with `?? 'Anonymous'`. No `any` reaches the typed field.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
Shantanu-TAM: Fixed in `e2a750c`. Removed `DEFAULT_CANVAS_WIDTH/HEIGHT` and import `PIP_CANVAS_WIDTH/HEIGHT` from `pipUtils` (already a dependency), so width/height can't drift.

packages/roomkit-react/src/Prebuilt/components/PIP/PIPManager.ts
Shantanu-TAM: Fixed in `e2a750c`. Removed `PIPChatMessage`; `PIPManager` now imports and uses the shared `PIPIncomingMessage` from `pipMessageUtils` (single producer→consumer type).

packages/roomkit-react/src/Prebuilt/components/PIP/pipUtils.js
Shantanu-TAM: Fixed in `e2a750c`. Extracted and exported `truncateToWidth` with an injectable `measure(str) => number` (no canvas needed to test) and switched the per-char shrink to a binary search (O(log n) measures). Added unit tests covering fit/empty/truncate/boundary cases.

raviteja83 (approved):
Re-reviewed (correctness / error-handling / tests). All prior-round critical findings resolved in e2a750c (renderLoop try/catch+finally, subscription error handling, type narrowing, dedup'd canvas constants + PIPChatMessage type, truncateToWidth extracted). Feature logic verified correct. CI green incl. dashboard-app.

Verified empirically that pip.test.js is excluded from the published bundle (not in esbuild's input graph; no pip.test.d.ts from tsc) — so the orphaned suite ships nothing.

Follow-ups (non-blocking): (1) wire a roomkit-react test runner so pip.test.js actually executes, + add `**/*.test.*` to tsconfig exclude; (2) pre-existing listenToStateChange listener-removal gap.

PR #3728, opened 2026-08-07
fix: recover the mic and camera after an OS interruption, and stop backgrounding from leaving the room

raviteja83:
An OS audio/video interruption — an incoming call, another app taking the device, safari backgrounded on iOS — left the mic and the camera silently dead, and the app had no way to know. Repro: join on iOS, answer a WhatsApp call, end it, come back.

## Audio

**Recovery was decided from the native track flags.** `shouldReacquireTrack` checks `readyState` and `muted`, and after an iOS interruption both lie: the track comes back `live` and unmuted with its capture unit stopped. So the native unmute path called `setEnabled`, found nothing to reacquire and left a silent mic, and the `visibilitychange` path early-returned before doing anything.

An interruption is now remembered, and both triggers — the native unmute and returning to the foreground — run one shared routine that:

1. resumes the shared `AudioContext`. iOS suspends it for the duration of the interruption and does not resume it on its own. Plugins publish the destination node of that context, so with noise suppression on the track stays silent even after being reacquired;
2. reacquires the track with a fresh `getUserMedia`, the only thing that reliably restores capture;
3. re-publishes the enabled state. Interruption start sends `mute=true` to biz, on which every remote peer unsubscribes from this peer's audio and resubscribes only on the next update — the visibility path never sent it, so the room kept treating a recovered peer as muted;
4. resumes the remote audio playback the OS paused alongside capture, which the visibility path also skipped.

## Video

A WhatsApp video call takes the camera too, and video had the same hole — in one place worse. `HMSLocalVideoTrack.setEnabled` returns early when the value is unchanged, so the `setEnabled(this.enabled)` at the end of `handleTrackUnmuteNatively` could never do anything. The native unmute cleared the published mute and re-played the video elements while the camera stayed frozen, so remote peers saw a live tile showing a still frame. On desktop that is the only path there is — the `visibilitychange` handler is attached on mobile only.

Video now mirrors audio: the interruption is remembered, and the native unmute and the foreground return run one shared routine that re-publishes the enabled state and resumes the paused remote playback. The camera is reacquired when it has to be — the background path turned it off, or the track reports it is not publishing. When the OS restarts capture itself the track is live again and replaying the sinks is enough; a `getUserMedia` there would only add a few hundred ms of frozen tile.

`localAudioUnmutedNatively` → `HMSVideoTrack.handleTrackUnmute` is unchanged. It re-triggers `play()` on the video elements, which recovers rendering, not capture.

## Both

Recovery is deferred while the page is hidden **on handheld platforms only**, where the OS does not hand capture back to a backgrounded tab and the foreground return retries. A hidden desktop tab is still a tab that can call `getUserMedia`, so deferring there would leave the mic dead — and the peer published as muted — until the user happened to come back to it.

That check is `isMobileOrTablet`, not `isMobile`: the latter is a device-type check, so it is false for every tablet, and for iPadOS 13+ safari it is false twice over — that sends a desktop-class user agent which parses as macOS with no device type at all. iPads are detected by the touch points a mac does not report. `isMobile` is left alone, since its other caller asks a different question.

Recovery is single-flighted, since both triggers fire on return and a second `getUserMedia` would stop the track the first one installed. The trigger that loses waits for the one in flight rather than returning, so whatever runs afterwards sees a finished recovery and not a half-replaced track.

`HMSTrackInterruption` carries the track type and is forwarded to the update/preview listener as `onTrackInterruption`, with `TRACK_INTERRUPTION_START` / `TRACK_INTERRUPTION_END` notifications for `useHMSNotifications`. It is exported from the entrypoint alongside `HMSTranscriptionInfo`, so an app that receives one of those notifications can name its payload. One prebuilt dialog names the device it applies to.

The app hears about an interruption only when the device is really not capturing and the page is visible — the only state a prompt can be acted on. A hidden page says nothing: nobody is there to read it, the device comes back on return, and on the video side the camera being reported on is the one the background path turned off itself. Locking the screen goes through the same path: the camera is released, remote peers see the avatar, and it comes back on unlock without a prompt.

If the device does not come back, the foreground path raises it then, once — reading either the interruption it recorded or the track flags, since neither is sufficient alone. The flags lie after an iOS interruption, and both video failure paths leave a track that reports `live` and unmuted, since backgrounding swaps in a blank canvas track and a failed reacquire installs one too. The flags are the only witness for a device that died without ever firing `mute`, and for a reacquire that resolved but handed back a muted camera. A revoked permission is raised on return too — it is the one state that cannot recover on its own.

The end is published wherever capture is actually restored, not only from the recovery routine, so a user who gets the device back another way — unmuting themselves reacquires the track without going near recovery — is not left with a prompt over a working device. It is published only once the track that was installed says it is capturing: `getUserMedia` resolving is not proof of that, since iOS can hand back a track that is already muted. Starts and ends stay paired: no end for a start that was never published.

Nothing is taken from a device that is not publishing, so an interruption is only recorded for an enabled track. A peer that joined muted holds an empty track, and `isEmptyTrack` stays true for it for the whole session, so anything keyed on `shouldReacquireTrack` alone would treat them as permanently interrupted and prompt them for a mic they never turned on.

The `interruption.start` / `interruption.stop` analytics events are unchanged and ungated: they record every interruption, including the ones the user never had to see.

## Backgrounding

On mobile there was often nothing left to recover, because backgrounding the browser could leave the room outright. `leaveOnUnload` listened for `freeze`, which chromium fires at the document of a hidden page on android, and for `pagehide` without looking at `persisted` — safari page-caches a document holding an `RTCPeerConnection`, so backgrounding on iOS fires it with `persisted === true`. Both of those pages resume, and the leave goes out as a user request over signalling, so the peer was removed for good with no reconnect.

The leave now runs only when the document is really going away: on `pagehide` that is not entering the cache, plus `beforeunload` for safari on desktop, which dispatches no `pagehide` when a tab is closed — and in both cases only while the page is still visible, which on a real unload it is, since the unload steps fire `pagehide` before flipping the visibility state to hidden. Closing a tab from the tab tray still leaves. `freeze` is no longer listened for at all. A page that is hidden, cached or frozen and never comes back is left to the server, which removes the peer once signalling stops. Reloading a visible page — including android pull-to-refresh — is a real unload and still leaves.

Also: audio output routing is logged (whether `setSinkId` was skipped and why, with browser and version — it is skipped on non-chromium, including safari 18.4+ where it causes robotic audio on bluetooth changes), and the playlist `AudioContextManager` resumes from safari's non-standard `'interrupted'` state, which its `=== 'suspended'` check missed.

24 tests across the two tracks cover reacquisition despite healthy-looking flags, recovery from the foreground event alone, the enabled-state republish, remote playback resume, the mobile hidden-page deferral and the desktop path that does not defer, single-flighting, no end-notification when recovery fails or when the reacquired device is still not capturing, the prompt when the device did not come back, the end when the user recovers the device themselves, silence for a device that was only backgrounded, silence for a peer that joined muted, and an already-off camera left alone.


sapta100ms (approved):
Tested ✅

PR #3736, opened 2026-08-20
fix: stop websocket reconnects from re-enabling noise cancellation against template policy

cyril-k-031225:
## Problem

`room.isNoiseCancellationEnabled` had two writers, and one of them was destructive.

**Writer 1** — `transport/index.ts:1114`, in `internalConnect`, writes the raw `noiseCancellation` `/init` feature flag.

**Writer 2** — `sdk/store/Store.ts`, in `handleNoiseCancellationPlugin`, ANDed the template policy decision against **its own prior value** rather than against the init flag:

```ts
this.room.isNoiseCancellationEnabled = !!plugin?.enabled && !!this.room.isNoiseCancellationEnabled;
```

Once writer 1 overwrote the field, writer 2 had lost the information it needed and could not recover the correct value.

`retrySignalDisconnectTask` calls `internalConnect` on **every websocket reconnect**, so each reconnect reset the field to the raw init flag. Policy re-arrived and corrected it, but not instantly — a measured **158–483 ms window** (4 runs) in which the Krisp gate was open.

### Impact

A customer disabled noise cancellation in their template on June 8. It took effect for 99.6% of joins. The remaining 0.4% kept Krisp running and kept billing 10–50 min/day for 10 weeks.

Only 0.4% because attaching Krisp needs a local audio track to already exist when the gate reopens. On first connect the track is created *after* policy, so the window is unreachable; once preview completes the track persists and every reconnect becomes a live opportunity. That combination is common on flaky mobile — all 44 leaking peers were mobile.

Two amplifiers made a 300 ms race expensive. First, the gate is checked only at add time, so nothing detaches Krisp once it is attached — fleet-wide `krisp.stop` fires at roughly 60% of the `krisp.start` rate, but it is ~0 for some tenants, and where it is, an attach lasts the life of the track. Second, billing charges the whole session, turning one peer × 19.5 min of exposure into 4 peers × 53.6 min billed (~2.8×).

Reproduced on a second account, session `6a8620b8d7913cfe895a014f` — Krisp running with `isNoiseCancellationEnabled: false`.

## Fix

The single mutable field becomes two independent source fields plus a get-only accessor:

```ts
isNoiseCancellationEnabledFromInit = false;    // written only by HMSTransport
isNoiseCancellationEnabledFromPolicy = false;  // written only by Store

get isNoiseCancellationEnabled(): boolean {
  return this.isNoiseCancellationEnabledFromInit && this.isNoiseCancellationEnabledFromPolicy;
}
```

Three properties follow:

1. **Both writes are idempotent.** No interleaving of init, policy, and reconnect can yield `true` unless both sources genuinely say `true`.
2. **The window is no longer exploitable.** Both sources default to `false`, so the pre-policy state denies instead of allows. The 158–483 ms interval still exists in time; it no longer grants anything.
3. **A third destructive writer can't be added.** Assigning to the accessor is a compile error.

Separately, `handleNoiseCancellationPlugin` is now called unconditionally from `setKnownRoles` instead of from the key-driven `switch`, so an absent `noiseCancellation` key resolves to disabled rather than leaving the raw init flag in place. Whiteboard and transcription stay in the switch, where absence correctly means no-op — they grant permissions additively.

**This one is a defensive invariant, not a live leak.** The server always emits the key: in the policy serializer `noiseCancellation` is a non-pointer struct field, so its `omitempty` is a no-op in Go, and the payload is always `"noiseCancellation":{}` (disabled) or `{"enabled":true}`. The enclosing `plugins` object is likewise always non-nil at its single construction site, which every connect goes through. So no template reaches the absent-key path today, and this change guards against a future serializer change rather than fixing an observed defect. Thanks @raviteja83 for pinning this down.

## Behavior change

None. The absent-key case above is unreachable from the server, so no live template loses noise cancellation as a result of this PR, and no template census is needed before shipping.

## Public API

Not affected. The two source fields live on the internal `Room` class, which is not re-exported from the package barrel; consumers get `HMSRoom` from `./schema`, which is unchanged. `interfaces/room.ts` gains a `readonly` marker, and that file is not part of the public export surface either. Verified against the built `dist/index.d.ts`.

## Tests

Four tests in `Store.noiseCancellation.test.ts`. They drive the init source through the real seam — `InitService.fetchInitConfig` → `isFlagEnabled(InitFlags.FLAG_NOISE_CANCELLATION)` → `internalConnect` — and read only the public `isNoiseCancellationEnabled`. They never name the two new source fields, so the file compiles against the pre-fix sources and fails there.

Regression tests, both **fail on `origin/main`** with the four source files reverted (`2 failed, 2 passed`):

- **`stays disabled when a reconnect re-writes the init flag after policy said no`** — the race. Pre-fix, the second `internalConnect` re-writes the field the policy writer owns → `true`.
- **`revokes the policy source when a later policy drops the noiseCancellation key`** — pre-fix, key-driven dispatch skips the writer entirely, so the earlier enable survives → `true`.

Mutation-killers, each verified to fail against a targeted mutant of the fixed code:

- **`stays disabled when the account flag is off even if the template enables it`** — deleting `isNoiseCancellationEnabledFromInit &&` from the getter fails only this test.
- **`is enabled only when both the account flag and the template policy allow it`** — guards the positive case, so the getter can't be mutated to a constant `false`.

Restoring the key-driven `switch` fails only the revocation test, so the two discriminate cleanly rather than overlapping.

`packages/hms-video-store`: **32 suites, 237 tests, all passing.** Lint and `tsc` clean.

## Deliberately out of scope

- **Revocation.** Nothing removes an already-attached Krisp when the derived value goes false. This PR prevents new attaches; a peer that attached during the pre-fix window keeps it until the track or session ends.
- **Reactive-store staleness.** A mid-session policy frame that flips the policy source doesn't refresh the reactive-store copy: `syncRoomState` never calls `convertRoom`, and `onRoomUpdate` fires only for peer-count / recording / RTMP / HLS / transcription. So `selectRoom().isNoiseCancellationEnabled` can stay stale-`true` while the SDK getter reads `false`. Pre-existing, and harmless in Prebuilt — the SDK gate is authoritative for the attach and the UI self-heals on the next room update — but this PR makes it newly reachable. Follow-up.
- **Roomkit auto-arm.** `AudioVideoToggle.tsx:335` re-adds Krisp on audio-track churn whenever the room gate and layout appdata both allow. Unchanged — it is the mechanism that converted an open gate into an attach, but the gate is now closed.
- **Layout and dashboard.** `noise_cancellation.enabled_by_default` stays `true`, and the dashboard switch still renders disabled-and-unchecked once the template flag is off — invisible and unclearable. Separate repos.
- **Immediate mitigation.** Killing the `noiseCancellation` Unleash flag for the affected account is an ops action independent of this branch.


raviteja83 (dismissed):

packages/hms-video-store/src/sdk/store/Store.noiseCancellation.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    room.isNoiseCancellationEnabledFromInit = true;
+    // Writer 2, policy arrives and disables.
+    store.setKnownRoles(policyWith({ [Plugins.NOISE_CANCELLATION]: { enabled: false } }));
+    expect(room.isNoiseCancellationEnabled).toBe(false);
+
+    // Writer 1 again — this is what retrySignalDisconnectTask -> internalConnect does
+    // on every websocket reconnect. When the policy writer ANDed against its own prior
+    // value instead of its own field, this reopened the Krisp gate for 158-483ms until
+    // policy re-arrived, and a peer with a live audio track could attach inside it.
+    room.isNoiseCancellationEnabledFromInit = true;
```
raviteja83: This test is not load-bearing — it passes against the unfixed code. Reverting all four source files to `origin/main` and running this file gives `1 failed, 1 passed`, and the one that passes is this one.

Reason: it drives writer 1 through `isNoiseCancellationEnabledFromInit`, which doesn't exist pre-fix, so the assignment lands on an inert own-property and `isNoiseCancellationEnabled` is `undefined` — never `true`. The race is never constructed, so both assertions hold vacuously.

Drive the init flag through the real seam (`HMSTransport` / `isFlagEnabled(InitFlags.FLAG_NOISE_CANCELLATION)`) and it will pin the actual regression.

packages/hms-video-store/src/sdk/models/HMSRoom.ts
```diff
@@ -27,9 +27,28 @@ export default class Room implements HMSRoom {
...
+  isNoiseCancellationEnabledFromPolicy = false;
   translationConfig?: Record<HMSTranscriptionMode, { enabled: boolean; roleLanguages?: Record<string, string> }>;
 
+  /**
+   * Noise cancellation is available only when the account's feature flag and the
+   * template policy both allow it. Derived, never stored, so that a reconnect
+   * re-writing the init flag cannot clobber the template's decision.
+   */
+  get isNoiseCancellationEnabled(): boolean {
+    return this.isNoiseCancellationEnabledFromInit && this.isNoiseCancellationEnabledFromPolicy;
```
raviteja83, line 49: The init-flag conjunct is unpinned. Deleting `isNoiseCancellationEnabledFromInit &&` from this line — i.e. removing the account entitlement check entirely — leaves both new tests green (`2 passed`).

Every test sets the init source to `true`, so the init-false case never executes. One more case fixes it: init `false` + policy `enabled: true` → expect `false`.

packages/hms-video-store/src/sdk/store/Store.ts
```diff
@@ -224,6 +224,11 @@ class Store {
   setKnownRoles(params: PolicyParams) {
     this.knownRoles = params.known_roles;
     this.addPluginsToRoles(params.plugins);
+    // Resolved unconditionally, not from the key-driven switch inside
+    // addPluginsToRoles: a policy that omits noiseCancellation means "not
+    // configured", which means off. Driving it from key presence let such
+    // templates keep the raw /init flag forever.
+    this.handleNoiseCancellationPlugin(params.plugins?.[Plugins.NOISE_CANCELLATION]);
```
raviteja83, line 231: Worth keeping as a defensive invariant, but the leak it's described as fixing isn't reachable: the server always emits the `noiseCancellation` key. In the policy serializer it's a non-pointer struct field, so its `omitempty` is a no-op in Go — the payload is always `"noiseCancellation":{}` (disabled) or `{"enabled":true}`. The enclosing `plugins` object is likewise always non-nil at its single construction site, which every connect goes through.

So the "second leak, no race required" section and the "count how many live templates would lose noise cancellation" gate both resolve to zero — no census needed before shipping. Suggest rewording rather than blocking on it.

packages/hms-video-store/src/sdk/store/Store.ts
```diff
@@ -522,8 +523,9 @@ class Store {
     if (!this.room) {
       return;
     }
-    // it will be called again after internalConnect room initialization, even after network disconnection
-    this.room.isNoiseCancellationEnabled = !!plugin?.enabled && !!this.room.isNoiseCancellationEnabled;
+    // Idempotent: writes only its own field, never reads the derived value. Called
+    // again on every policy notification, including after a reconnect.
+    this.room.isNoiseCancellationEnabledFromPolicy = !!plugin?.enabled;
```
raviteja83, line 528: Minor, and pre-existing, but this PR makes it newly reachable: a mid-session policy frame that flips this source doesn't refresh the reactive-store copy. `syncRoomState` never calls `convertRoom`, and `onRoomUpdate` fires only for peer-count / recording / RTMP / HLS / transcription — so `selectRoom().isNoiseCancellationEnabled` can stay stale-true while this getter reads `false`.

Harmless in Prebuilt (the SDK gate is authoritative for the attach, and the UI self-heals on the next room update), but worth a line in the description or a follow-up.

raviteja83 (commented):
Design is right — the two-source + derived-getter split closes the reconnect race structurally, and I verified there is no other write path, that `Store`/`Room` survive preview→join and every reconnect (so the policy source is never reset), that `convertRoom` reads the getter by name (no prototype-accessor drop), and that the public export surface is unchanged.

Two things to fix before merge, both in the tests: test 1 passes against the unfixed code, and the init-flag conjunct in the getter is unpinned. Details inline.

Two claims in the description also need correcting: the "absent `noiseCancellation` key" case is unreachable from the server, so the template census isn't needed; and `krisp.stop` is not ~0 in aggregate (it fires at roughly 60% of the `krisp.start` rate overall, though it really is ~0 for some tenants).

raviteja83 (approved):

PR #3742, opened 2026-08-25
fix(store): stop a replaced subscription request from replaying stale state

raviteja83:
A retry replays the bytes serialised when the request was made, so a request still retrying after a newer one has been made for the same track puts the older value back — a peer unmutes and goes silent again ten seconds later, or a tile resizes and snaps back to the layer it had. Most of these calls are fire-and-forget, so nothing surfaces when it happens.

The subscribe connection now tracks the newest request per `method` + `track_id`, and a request that no longer holds that claim stops retrying. Being replaced is not a failure: it resolves the way the existing `disableAutoUnsubscribe` skip does, rather than reporting an error no caller can act on, and the check also guards the out-of-retries throw so a request replaced on its last attempt does not report failure either.

Also catches the two `setEnabled` calls in `TrackManager`. A remote peer muting or unmuting applies the subscription over the same data channel, and both call sites discarded the promise, so an unanswered request surfaced as an unhandled rejection — in node, that takes the process down. The local mute state is already applied by then; only the request failed.


cyril-k-031225:
### Code review

Found 2 issues:

1. The `superseded()` check is missing on the third exit path. `sendMessage` evaluates it at the top of each retry attempt and in the final `if (!response)` block, but not when the SFU answers with a non-retryable, non-404 error code — that branch throws unconditionally. So a request already replaced by a newer one for the same track still rejects, which is exactly what the comment three lines above says it should not do ("not a failure - the newer request owns the outcome, so resolve ... rather than reporting an error nobody can act on"). Impact is bounded: `HMSRemoteVideoTrack.requestLayer` already catches, and `TrackManager.logIfRejected` now catches, so in practice this surfaces as a spurious error-level "could not apply mute state" log for a case this PR declares is not a failure — plus a live rejection at any remote-audio `setEnabled` caller that doesn't wrap the promise. Not covered by `staleSubscriptionState.test.ts`, which only exercises the never-answered path via `exhaustRetries`.

https://github.com/100mslive/web-sdks/blob/b325914c07c22a69f2a221e3333f68e77a9b4c09/packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts#L227-L231

2. The "no unhandled rejection leaked" assertion is inert. jsdom never dispatches a DOM `unhandledrejection` event for a real unhandled rejection — in jsdom 20.0.3 the name appears only in the event-handler IDL attribute list in `lib/jsdom/browser/Window.js`, with no dispatch anywhere — so `rejections` is always `[]` and `expect(rejections).toHaveLength(0)` passes identically with or without `logIfRejected`. Confirmed by running it: register the listener, `Promise.reject(...)`, zero events. The test still earns its keep via `expect(logError).toHaveBeenCalled()`; it's the leak half that proves nothing. This is the same pathology [5d0a4d5c6](https://github.com/100mslive/web-sdks/commit/5d0a4d5c65bd31cdf9ece77e081e030dd7d9b676) fixed a day earlier one layer down — "replace an unhandledRejection spy that could never fire - jest installs its own handler, so the assertion was inert" — reintroduced with `window` in place of `process`.

https://github.com/100mslive/web-sdks/blob/b325914c07c22a69f2a221e3333f68e77a9b4c09/packages/hms-video-store/src/notification-manager/managers/trackMuteFailure.test.ts#L54-L69

Checked and did not flag: the `latestRequestPerState` set/delete lifecycle and the `finally` guard (a superseded request never deletes the newer entry); `{ id: requestId } as PreferLayerResponse` against callers (`PreferLayerResponse` has only `id` and optional `error`, and no caller reads either); local state consistency (`setAudio`/`setVideoLayerLocally` apply optimistically before the network call); the per-track key (one track's churn cannot cancel another's); and the `disableAutoUnsubscribe` skip, which returns before the map is touched.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cyril-k-031225:
### Code review

Re-reviewed at `b01dd8f`. Both issues resolved, no new ones.

1. Supersession now agrees across every exit from the retry loop. `dropped()` is shared by all three failure paths, and the non-retryable branch checks `superseded()` before throwing. Walking the exits: loop-top, attempt-threw, non-retryable, retryable-sleep, retries-exhausted — all consistent now; the 404 `break` returns its response as before, unrelated to supersession. The two new tests pin both directions, which is the part that was missing.

https://github.com/100mslive/web-sdks/blob/b01dd8faa1348cd150cb9e2789f4aca28cc7ca1d/packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts#L231-L239

2. The inert assertion is gone, and replaced with one that can actually fail. `jest.spyOn(rejection, 'catch')` calls through, so `logError` still exercises the real path, and the handler attaches in the same synchronous turn as `handleTrackUpdate` — so the seeded rejection never leaks for real either. The JSDoc recording why the jsdom listener proved nothing is a good thing to leave behind.

https://github.com/100mslive/web-sdks/blob/b01dd8faa1348cd150cb9e2789f4aca28cc7ca1d/packages/hms-video-store/src/notification-manager/managers/trackMuteFailure.test.ts#L49-L71

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cyril-k-031225 (approved):

PR #3743, opened 2026-08-26
fix(store): report subscribe request failures correctly

raviteja83:
**1 of 3.** One file plus its tests. Independent of the other two.

The bounded waits added for the silent-recording fix turned a lost SFU reply from a hang into a rejection. Four things in that path then report the wrong outcome:

- A retryable error arriving on the **final** attempt was returned as the resolved value. No caller inspects `error` on a resolved response, so a request the SFU refused read as one it applied.
- `error.code / 100 === 5` is exact division and only ever matched `500` — a `502`/`503`/`504` from a restarting SFU threw on the first attempt with no retry at all.
- The supersession claim was checked *before* the channel-open wait but not after, so a request replaced while parked still put stale desired state on the wire. For `prefer-audio-track-state` that is an audible unsubscribe nobody asked for.
- `close()` left the retry loops running against a dead channel, and the eventemitter2 waits subscribed until their own timeouts on an emitter capped at 60 listeners. Pending requests now settle at once and cancel their waits — reported as *dropped*, not as failures: a leave is not a failure, and nothing survives `clearPeerConnections()` to act on the state. One caveat: the retry backoff is the one wait `close()` cannot cut short, since `workerSleep` has no cancel.

The retry backoff no longer runs after the final attempt, where it only delayed the throw and could not be cut short by `close()`.

On exit: a response the SFU actually sent is the outcome whatever happened to the claim meanwhile, and a superseded request is reported as dropped rather than as a failure — otherwise every normal leave logs an error for requests the code itself declares are not failures.

## Scope
Deliberately just the retry loop. An earlier version of this PR also reworked `HMSRemoteStream` so that a failed request could not poison the equality guards that suppress no-op requests — 65 lines, three fields and two new gates for it. Review found it introduced more failure modes than it removed (a counter that overshot and permanently re-armed the guard it was bypassing; no ground-truth reset on the audio side). That problem is real and still open, but it needs the desired-vs-confirmed split rather than a bypass bolted onto optimistic state, so it is not in this PR.

## Test plan
17 tests. **10 of them fail against `main`**; the other seven are regression guards for the restructure rather than verified-failing bug tests, and are not counted as such. Every production hunk is mutation-checked. 39 suites / 300 tests pass on this branch alone.

## Compatibility with 0.14.7
No behavioural change on the success path. Everything here is reached only when a subscribe request fails, which in 0.14.7 could not happen — that release had no timeouts and the caller hung instead.

### DIFF
--- a/packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
+++ b/packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
@@ -23,25 +23,37 @@ describe('HMSSubscribeConnection api data channel', () => {
 
   beforeEach(() => {
     sent = [];
-    window.RTCPeerConnection = jest.fn().mockImplementation(() => ({
-      close: jest.fn(),
-      getSenders: () => [],
-    })) as unknown as typeof RTCPeerConnection;
+    window.RTCPeerConnection = jest
+      .fn()
+      .mockImplementation(() => ({ close: jest.fn() })) as unknown as typeof RTCPeerConnection;
     const signal = { trickle: jest.fn() } as unknown as JsonRpcSignal;
     const observer = { onApiChannelMessage: jest.fn() } as unknown as ISubscribeConnectionObserver;
     connection = new HMSSubscribeConnection(signal, {}, () => false, observer);
 
+    let readyState = 'open';
     const nativeChannel = {
       label: API_DATA_CHANNEL,
-      readyState: 'open',
-      send: (message: string) => sent.push(message),
-      close: jest.fn(),
+      get readyState() {
+        return readyState;
+      },
+      send: (message: string) => {
+        if (readyState !== 'open') {
+          throw Error('InvalidStateError: channel is not open');
+        }
+        sent.push(message);
+      },
+      close: jest.fn(() => {
+        readyState = 'closed';
+      }),
     } as unknown as RTCDataChannel;
     connection.nativeConnection.ondatachannel?.({ channel: nativeChannel } as RTCDataChannelEvent);
   });
 
-  // a test that throws before its inline useRealTimers() would leak fake timers into the next one
   afterEach(() => {
+    // settles anything still parked, so a discarded promise cannot reject minutes later attributed
+    // to whatever suite is running by then
+    connection.close();
+    // a test that throws before its inline useRealTimers() would leak fake timers into the next one
     jest.useRealTimers();
   });
 
@@ -256,9 +268,10 @@ describe('HMSSubscribeConnection api data channel', () => {
     await jest.advanceTimersByTimeAsync(100);
 
     expect(settled).toBe(true);
-    // rejects rather than resolving: a caller that reads a torn-down connection as success keeps
-    // believing the SFU applied state it never received
-    await expect(promise).resolves.toBeInstanceOf(Error);
+    // dropped, not rejected: a leave is not a failure, and nothing survives close() to be wrong
+    // about - throwing here wrote one error per in-flight track on every normal leave
+    await expect(promise).resolves.not.toBeInstanceOf(Error);
+    await expect(promise).resolves.not.toHaveProperty('error');
   }, 10_000);
 
   /** the backoff before an attempt that will never happen only delays the throw */
@@ -296,11 +309,15 @@ describe('HMSSubscribeConnection api data channel', () => {
       params: { max_spatial_layer: 'high', track_id: 'track-1' },
     });
     await Promise.resolve();
-    // a newer request for the same track takes the claim while the first reply is in flight
-    connection.sendOverApiDataChannelWithResponse({
-      method: 'prefer-video-track-state',
-      params: { max_spatial_layer: 'low', track_id: 'track-1' },
-    });
+    // a newer request for the same track takes the claim while the first reply is in flight.
+    // Nothing answers it and this test runs on real timers, so its rejection needs a handler or it
+    // lands ~30s later attributed to whatever suite is running then.
+    connection
+      .sendOverApiDataChannelWithResponse({
+        method: 'prefer-video-track-state',
+        params: { max_spatial_layer: 'low', track_id: 'track-1' },
+      })
+      .catch(() => undefined);
     await Promise.resolve();
     emitReply(sent[0], { result: { track_id: 'track-1' } });
 
@@ -331,9 +348,13 @@ describe('HMSSubscribeConnection api data channel', () => {
     connection.close();
     await jest.advanceTimersByTimeAsync(100);
 
-    await expect(stale).resolves.not.toBeInstanceOf(Error);
-    await expect(stale).resolves.not.toHaveProperty('error');
-    await expect(latest).resolves.toBeInstanceOf(Error);
+    // both drop: one because it was superseded, the other because close() is the same category.
+    // Asserting on `latest` too rather than merely awaiting it - otherwise it reads as though it
+    // is only there to keep a rejection handled.
+    for (const settled of [stale, latest]) {
+      await expect(settled).resolves.not.toBeInstanceOf(Error);
+      await expect(settled).resolves.not.toHaveProperty('error');
+    }
   }, 10_000);
 
   /**
@@ -427,7 +448,7 @@ describe('HMSSubscribeConnection api data channel', () => {
     await jest.advanceTimersByTimeAsync(100);
 
     expect(settled).toBe(true);
-    await expect(promise).resolves.toBeInstanceOf(Error);
+    await expect(promise).resolves.not.toHaveProperty('error');
   }, 10_000);
 
   /**
@@ -479,7 +500,7 @@ describe('HMSSubscribeConnection api data channel', () => {
     await jest.advanceTimersByTimeAsync(100);
 
     expect(settled).toBe(true);
-    await expect(promise).resolves.toBeInstanceOf(Error);
+    await expect(promise).resolves.not.toHaveProperty('error');
   }, 10_000);
 
   /** a channel that opens after the first attempt gave up should still get the request through */
--- a/packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
+++ b/packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
@@ -158,7 +158,7 @@ describe('HMSSubscribeConnection api data channel', () => {
     await jest.advanceTimersByTimeAsync(10);
 
     expect(sent).toHaveLength(3);
-    await expect(promise).resolves.toBeInstanceOf(Error);
+    await expect(promise).resolves.toMatchObject({ message: expect.stringContaining('code=500') });
   }, 10_000);
 
   /**
@@ -332,6 +332,7 @@ describe('HMSSubscribeConnection api data channel', () => {
     await jest.advanceTimersByTimeAsync(100);
 
     await expect(stale).resolves.not.toBeInstanceOf(Error);
+    await expect(stale).resolves.not.toHaveProperty('error');
     await expect(latest).resolves.toBeInstanceOf(Error);
   }, 10_000);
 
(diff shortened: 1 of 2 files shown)


packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
```diff
@@ -213,43 +246,76 @@ export default class HMSSubscribeConnection extends HMSConnection {
...
+        const shouldRetry = Math.floor(error.code / 100) === 5 || error.code === 429;
         if (!shouldRetry) {
           if (superseded()) {
             return dropped();
           }
           throw Error(`code=${error.code}, message=${error.message}`);
         }
-        const delay = (2 + Math.random() * 2) * 1000;
-        await workerSleep(delay);
+        if (i < this.MAX_RETRIES - 1) {
```
cyril-k-031225, line 293: `close()` still can't cut this backoff short, so "pending requests now reject promptly" holds only for the two `waitFor` parks. A request sleeping here when `close()` lands keeps sleeping for the remaining 2–4s, then wakes, hits the loop-top `if (this.closed) break`, and throws. Not a correctness problem, but it's the one wait on this path that the new abort mechanism doesn't cover.

Worth noting too that `workerSleep` spawns a fresh `Worker` (plus an un-revoked `createObjectURL` blob) per call and only `terminate()`s on tick — so a leave during backoff leaves one live worker per pending track until it fires. Dropping the post-final-attempt sleep already removes one of those per failing request, which is nice.

If you want the sleep abortable, it fits the same shape you already built:

```ts
const cancelableSleep = workerSleep(delay) as CancelablePromise<void>;
await this.abortOnClose(cancelableSleep);
```

though `workerSleep` has no `cancel`, so `abortOnClose` would settle the race while the worker keeps running to its tick — you'd need `workerSleep` to hand back a terminator for the leak half. Fine to leave as-is and just say so; flagging because the PR description claims promptness.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
```diff
@@ -213,43 +246,76 @@ export default class HMSSubscribeConnection extends HMSConnection {
...
     }
-    return response;
+    if (response) {
+      return response;
+    }
+    if (superseded()) {
+      return dropped();
+    }
+    if (this.closed) {
+      throw Error(`Subscribe connection closed before ${requestId} was answered - ${request}`);
```
cyril-k-031225: This is the exit I'd push back on, by the PR's own rule.

The stated principle is "a request the newer one took over is not a failure" — because otherwise every normal leave logs an error for something the code itself declares isn't one. A request aborted by `close()` is in exactly that category: `clearPeerConnections()` runs on `leave()` and on `handleSFUMigration()`, and neither is a failure the app or the caller can act on. But this exit throws, and the video path routes it straight to an error log:

`requestLayer` rethrows → `addSink`/`removeSink` reject → `VideoElementManager.logIfRejected` → `HMSLogger.e('addSink failed', ...)`, deliberately at `e` "because a warn is dropped once an app calls setLogLevel(ERROR)". So every leave with in-flight `prefer-video-track-state` requests now writes one error per track.

To be fair this isn't a regression — before the PR those same requests burned their retry budget against the dead channel and threw `No response from SFU` ~30s later, logging the same thing. The change is that it's now immediate and reliable rather than delayed, which makes it much more visible on a normal leave.

The test comment argues the opposite side — "a caller that reads a torn-down connection as success keeps believing the SFU applied state it never received" — and that's a reasonable position. But nothing downstream distinguishes this rejection from a real SFU refusal, so the cost of being right is unconditional error-level noise on the happy path. Three options, in the order I'd prefer them:

1. Treat `closed` like `superseded` and `dropped()` it. Consistent with the disableAutoUnsubscribe skip, and nothing survives a close to act on the stale belief anyway — `clearPeerConnections()` nulls the connection, and SFU migration also drops every `HMSRemoteStream` via `removeRemoteTracks()`, so the optimistic `audio`/`video` fields go with it.
2. Keep the throw but make it identifiable (a typed error / an `isClose` flag) so `logIfRejected` can log it at debug.
3. Keep it as-is and accept the noise — but then it'd be worth saying so in the PR description, since it reads today as though the leave-time error logs are gone.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
```diff
@@ -181,10 +191,33 @@ export default class HMSSubscribeConnection extends HMSConnection {
...
+   * Settles `wait` early if the connection is closed while it is parked on the api data channel.
+   * The wait is cancelled rather than just abandoned: eventemitter2 keeps the listener subscribed
+   * until its own timeout otherwise, and the emitter is capped at 60.
+   */
+  private abortOnClose = async <T>(wait: CancelablePromise<T>): Promise<T> => {
+    let abort!: (error: Error) => void;
+    const aborted = new Promise<never>((_, reject) => {
+      abort = reject;
+    });
+    this.pendingAborts.add(abort);
```
cyril-k-031225, line 217: I went and checked the eventemitter2 claim in the doc comment, since the whole cancel path rests on an untyped property. It holds, for `^6.4.9`: `waitFor` goes through `makeCancelablePromise`, and because native `Promise.prototype.cancel` doesn't exist, `isCancelable` is false and the `promise.cancel` assignment on the `!isCancelable` branch does happen. Calling it runs the `onCancel` callback, which is `self.off(event, listener)` — so the listener really is unsubscribed, not just abandoned. Worth pinning `eventemitter2` to `~6.4.x` at some point given the dependency on undeclared behaviour, but it's correct today.

Two things fall out of reading that source, both fine as written — noting them so the next reader doesn't have to re-derive them:

- `cancel()` also **rejects** `wait` (`_reject(reason || Error('canceled'))`). On the abort path that rejection arrives after `Promise.race` has already settled, and it's harmless only because `race` subscribed to `wait` — which is what your comment above is about. Good call putting it there.
- On the success and timeout paths `cleanup()` has already nulled `callbacks`, so the `finally`'s `cancel()` is a no-op rather than a double-reject.

One hardening suggestion: register the abort only after re-checking `closed`.

```ts
if (this.closed) {
  wait.cancel?.();
  throw Error('Subscribe connection closed');
}
this.pendingAborts.add(abort);
```

This is unreachable today — I traced it, and there's no `await` between the `if (this.closed) break` at the top of the loop and `pendingAborts.add`, nor between the aborted-wait rejection and the `break` in the catch, so a wait can't be registered after `close()` has already drained and `clear()`ed the set. But it's one refactor away from being reachable, and the failure mode is silent: an abort in a set nobody will ever iterate, and a wait that sits for the full `RESPONSE_TIMEOUT` on a connection that's gone.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
```diff
@@ -213,43 +246,76 @@ export default class HMSSubscribeConnection extends HMSConnection {
...
+     * discarding it would report a request the SFU applied as dropped.
+     */
+    if (response?.error) {
       if (superseded()) {
         return dropped();
       }
-      throw Error(`No response from SFU for ${requestId} after ${this.MAX_RETRIES} tries - ${request}`);
+      // the loop can run out of attempts still holding a retryable error, and no caller inspects
+      // `error` on a resolved response, so returning it here reads as a request the SFU applied
+      throw Error(`code=${response.error.code}, message=${response.error.message}`);
```
cyril-k-031225: This message is byte-identical to the non-retryable throw ~25 lines up, so the two most diagnostically different outcomes are indistinguishable in a log: "the SFU refused this once and told us not to retry" vs "the SFU failed it three times in a row". The timeout exit right below keeps its `after ${this.MAX_RETRIES} tries` — this one drops it for no reason I can see.

```suggestion
      throw Error(`code=${response.error.code}, message=${response.error.message} after ${this.MAX_RETRIES} tries`);
```

Cheap, and this PR exists because the wrong outcome reached a log during an incident.

Minor, on line 280 while I'm here: `Math.floor(error.code / 100) === 5` is the right fix, but `error.code >= 500 && error.code < 600` states the intent without the reader having to check that the original was integer division gone wrong. Also worth knowing that a malformed error with no numeric `code` (`Math.floor(NaN) !== 5`) lands in the non-retryable branch and throws `code=undefined, message=...` — same as before this PR, so not yours to fix, just not obvious from the code.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
```diff
@@ -127,6 +136,242 @@ describe('HMSSubscribeConnection api data channel', () => {
...
+      .sendOverApiDataChannelWithResponse({
+        method: 'prefer-video-track-state',
+        params: { max_spatial_layer: 'low', track_id: 'track-1' },
+      })
+      .catch((error: Error) => error);
+    await jest.advanceTimersByTimeAsync(10);
+    emitReply(sent[2], { error: { code: 500, message: 'internal' } });
+    await jest.advanceTimersByTimeAsync(120_000);
+
+    await expect(stale).resolves.not.toBeInstanceOf(Error);
```
cyril-k-031225, line 396: **This assertion doesn't discriminate the behaviour it claims to test.** I checked out `main`'s `subscribeConnection.ts`, kept this PR's test file, and ran it — this test passes against `main`, for the opposite reason:

- On `main`: exhausted retryable error → `response` is set → falls past `if (!response)` → `return response`, i.e. resolves with `{ error: { code: 500 } }`. Not an `Error` instance → ✓
- On this branch: `superseded()` → `dropped()` → resolves with `{ id }`. Not an `Error` instance → ✓

`not.toBeInstanceOf(Error)` can't tell "dropped" from "returned the 500 as if it were the outcome", which is the exact confusion the PR is fixing. Assert on the shape instead:

```suggestion
    await expect(stale).resolves.not.toHaveProperty('error');
```

or tighter still, `resolves.toEqual({ id: expect.any(String) })` to pin it to the `dropped()` payload.

Same run says the mutation-check claim in the PR description needs a small correction: 10 of the 12 new tests fail against `main`, not 12. This one and `returns a response that arrived even if a newer request claimed the state meanwhile` both pass on `main` — the latter legitimately, since it's guarding a discard that only the new exit block could introduce, but it's a regression guard rather than a verified-failing test and the description shouldn't count it as one.

While here: `latest` on the next line is awaited but never asserted. If it's only there to keep the rejection handled, `await expect(latest).resolves.toBeInstanceOf(Error)` documents why it exists at no extra cost.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
```diff
@@ -127,6 +136,242 @@ describe('HMSSubscribeConnection api data channel', () => {
...
+    const first = connection.sendOverApiDataChannelWithResponse({
+      method: 'prefer-video-track-state',
+      params: { max_spatial_layer: 'high', track_id: 'track-1' },
+    });
+    await Promise.resolve();
+    // a newer request for the same track takes the claim while the first reply is in flight
+    connection.sendOverApiDataChannelWithResponse({
+      method: 'prefer-video-track-state',
+      params: { max_spatial_layer: 'low', track_id: 'track-1' },
+    });
```
cyril-k-031225: This second request's promise is discarded, and it's the one new test that doesn't install fake timers — so it's left parked on real `setTimeout`s. Nobody ever replies to it, so it runs its three 10s `waitForResponse` timeouts (the backoff is mocked to resolve immediately) and rejects with `No response from SFU` about 30s of wall clock later, with no handler attached.

Today that's latent rather than broken: the whole package suite finishes in ~8s locally (39 suites / 296 tests pass — the description says 295), so the process exits long before the timer fires. But it's a rejection whose timing is decoupled from any test, so if the suite ever grows past ~30s it surfaces attributed to whatever file happens to be running, which is a miserable thing to debug. Worth spending a character on:

```suggestion
    }).catch(() => {});
```

An `afterEach(() => connection.close())` alongside the new `useRealTimers()` one would cover this and any future variant — with this PR's changes that settles every parked request immediately, which is a nice side benefit of the abort work.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
```diff
@@ -32,10 +35,16 @@ describe('HMSSubscribeConnection api data channel', () => {
       label: API_DATA_CHANNEL,
       readyState: 'open',
       send: (message: string) => sent.push(message),
+      close: jest.fn(),
```
cyril-k-031225: `close` is stubbed but `readyState` is a fixed `'open'` literal, so after `connection.close()` this mock still claims the channel is open. The two close tests therefore only ever exercise the `this.closed` flag — `waitForChannelOpen`'s early return keeps firing, and `apiChannel.send()` keeps succeeding, on a channel the test just closed.

That's enough to pin the new flag, but it leaves the real post-close path untested: a genuine `RTCDataChannel` goes to `'closing'`/`'closed'`, at which point requests park on `waitFor('open')` (the wait the abort mechanism exists for) and `send()` throws into the `catch`. Backing it with a mutable field is a two-line change:

```ts
let readyState = 'open';
const nativeChannel = {
  label: API_DATA_CHANNEL,
  get readyState() {
    return readyState;
  },
  send: (message: string) => sent.push(message),
  close: jest.fn(() => {
    readyState = 'closed';
  }),
} as unknown as RTCDataChannel;
```

`settles pending requests when the connection is closed` should still pass, but via the abort rather than via the flag — which is what it's really meant to prove.

Unrelated: is `getSenders: () => []` on the peer-connection mock above load-bearing? `HMSConnection.close()` is just `this.nativeConnection.close()`, and I don't see another path from these tests into it. Tests pass without it if not.

---
_Generated by [Claude Code](https://claude.ai/code)_

cyril-k-031225 (commented):
Reviewed the diff and traced the callers. All four reported bugs are real and the fixes are right; the scope discipline is the best thing about this PR — the `HMSRemoteStream` rework named in the description genuinely doesn't belong here.

**What I verified locally** (installed the workspace, ran on `055faae`):

- `subscribeConnection.test.ts`: 13/13 pass in ~3.7s. Full package: **39 suites / 296 tests pass** (description says 295).
- Mutation check: I checked out `main`'s `subscribeConnection.ts`, kept this branch's test file, and re-ran. **10 of the 12 new tests fail against `main`**, not 12. Two pass — `returns a response that arrived even if a newer request claimed the state meanwhile` (fine; it guards a discard only the new exit block could introduce) and `drops rather than throwing when a superseded request exhausts its retries on an error` (**not fine** — it passes on `main` for the opposite reason, see the inline comment; that assertion needs tightening).
- The `eventemitter2` claim in the `abortOnClose` doc comment holds for `^6.4.9`: `waitFor` → `makeCancelablePromise`, native `Promise` has no `cancel`, so the `promise.cancel` assignment does happen and it calls `off(event, listener)`. The listener-leak fix is real, not a silent no-op on an undeclared property. Details inline.

**The four fixes.** The final-attempt error escaping as a resolved value, `error.code / 100 === 5` matching only `500`, the missing post-open-wait supersession check, and `close()` leaving the loops running — all confirmed by reading the code, and the reasoning about why each one matters downstream (`requestLayer` handing the response to `addSink`'s on-demand branch, `prefer-audio-track-state` producing an audible unsubscribe) checks out. Resetting `response = undefined` at the top of each iteration and returning early on 404 are both load-bearing given the new exit block; good that they're there.

I also confirmed the `closed` flag being permanent is safe: both `clearPeerConnections()` call sites (`leave()` and `handleSFUMigration()`) null the field, so the next join constructs a fresh instance and never reuses a closed one.

**Worth changing before merge:**

1. **The non-discriminating assertion** on the superseded-exhausts-retries test (inline). This is the one I'd insist on — it currently passes whether the code drops the request or hands back the 500 as the outcome.
2. **Should a close-aborted request throw?** By the PR's own "a request the newer one took over is not a failure" rule, a request killed by `leave()` isn't one either — and the video path routes this rejection to `HMSLogger.e` via `VideoElementManager.logIfRejected`, so every leave with in-flight layer requests writes one error per track. Not a regression (before, the same error arrived ~30s later via `No response from SFU`), but the description reads as though leave-time error logs are gone, and they aren't. Inline comment has three options.
3. **Smaller:** the exhausted-retry throw is byte-identical to the non-retryable one — losing "was this refused once or three times?"; `close()` still can't interrupt the 2–4s backoff, so "reject promptly" has an asterisk; the data-channel mock's `readyState` never leaves `'open'`, so the close tests exercise the flag rather than the real closed-channel path; one discarded promise left on real timers.

**Pre-existing, out of scope but adjacent** — `sendOverApiDataChannelWithResponse`'s optional `requestId` parameter (line 167) is a trap. It goes onto the wire as the request `id`, but `sendMessage` is called with the internal `id`, so `waitForResponse` filters for a uuid that never appears in any reply — any caller passing it would time out on all three attempts and land in exactly the wrong-outcome-reporting territory this PR is about. Nothing passes it today (only the two `HMSRemoteStream` call sites exist). Since this PR is already auditing this method's failure reporting, deleting the parameter would be a cheap way to close the hole; equally fine to leave for the desired-vs-confirmed follow-up.

The `Compatibility with 0.14.7` note is accurate — every path touched here is unreachable without the bounded waits, so the success path is unchanged.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
raviteja83: Took option 1 — it drops now. I checked your premise and it holds: `clearPeerConnections()` nulls `subscribeConnection`, and `handleSFUMigration()` additionally calls `store.removeRemoteTracks()`, so every `HMSRemoteStream` carrying the optimistic `audio`/`video` fields goes with it. There is nothing left to be wrong about, and one error per in-flight track on every leave is a real cost for being technically right.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
raviteja83: Left as-is and said so — the docblock now states the abort covers the two `waitFor` parks only, and that a request inside the backoff sleeps out the remaining 2–4s since `workerSleep` has no cancel. PR description updated too; it did read as though leave-time waits were all covered. The worker-per-call leak is worth its own issue rather than riding along here.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
raviteja83: Thanks for verifying it rather than taking the comment's word — the comment was wrong in the other direction and is gone. eventemitter2 does ship `CancelablePromise` in its `.d.ts` (the same file this imports `WaitForOptions` from); the local alias existed only because the library types `cancel` as taking a required reason while I called `cancel?.()`. Using the library type and passing a reason deleted the alias.

The re-check you suggested is already there — `abortOnClose` tests `this.closed` and cancels before `pendingAborts.add`. Agreed it is unreachable today and cheap insurance.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
raviteja83: Fixed, and you and my own mutation run landed on it independently. Worth noting `not.toHaveProperty('error')` alone is also insufficient — a regression that *throws* yields an `Error`, which has no `error` property either, so both halves are needed. Both are asserted now.

Description corrected: 10 of 17, not 12, with the regression guards called out as such. `latest` is asserted rather than just awaited.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
raviteja83: Both — a `.catch` on that promise, and the `afterEach(() => connection.close())` you suggested, which settles anything still parked and covers future variants.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.test.ts
raviteja83: Good catch — the mock tracks `readyState` now and `send()` throws once shut, so `settles pending requests when the connection is closed` goes through the abort rather than the flag. And no, `getSenders` was not load-bearing: removed, suite still green.

cyril-k-031225 (approved):

PR #3744, opened 2026-08-26
fix(store): apply audio volume to the element before it is attached and played

raviteja83:
**2 of 3.** Split out of the original omnibus PR. Independent of the other two.

Turning a peer down is two things: the audio element's `volume`, which is local and immediate, and the subscription state, which is a round trip to the SFU. `setVolume` awaited the round trip *before* touching the element, and the silent-recording fix then moved the whole call to after `srcObject` and `play()` so that attaching could not fail with it.

Together: **a peer added while the app volume is 0 was audible at full volume until the SFU answered — and forever if it never did**, since the assignment was never reached.

### What changed
- The local half runs synchronously, before the element is attached. `setAudioElement` carries the requested volume onto every element it is given, so a fresh element — including the one rebuilt during `MEDIA_ERR_DECODE` recovery — starts at the volume the user asked for rather than at `1`. Only the subscription half is left to the round trip, still un-awaited.
- `AudioSinkManager` records the new volume *before* the fan-out. It was assigned only once every track's round trip settled, so for that whole window every newly added track was given the volume the user had just turned off — and a slow older call could land after a newer one and overwrite it (drag the volume slider).
- `Store.updateAudioOutputVolume` fans the per-track requests out concurrently instead of awaiting them in series and abandoning the rest on the first rejection, which left later tracks on the old volume. Logged rather than thrown, unlike the `updateAudioOutputDevice` beneath it: `setVolume` turns the element down synchronously, so a rejection means the volume *is* applied everywhere and only the unsubscribe optimisation was lost. A sink change has no local half, which is why that one has to surface.

## Test plan
The decisive test asserts the element's volume **at the moment `play()` is called**, not just once everything settles — the earlier version of this fix passed a settle-time assertion while still being audible for the duration of `play()`. Verified failing against `main`; 41 suites / 294 tests pass on this branch alone.

## Compatibility with 0.14.7
`setVolume`'s observable contract is unchanged — same signature, same promise, same final state. Only the ordering within it moved, and it now applies the volume in strictly more cases than before (a rebuilt element used to silently revert to full volume).

### DIFF
--- a/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
+++ b/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
@@ -170,8 +170,8 @@ export class AudioSinkManager {
     audioEl.srcObject = new MediaStream([track.nativeTrack]);
     callListener && this.listener?.onTrackUpdate(HMSTrackUpdate.TRACK_ADDED, track, peer);
     await this.handleAutoplayError(track);
-    // the subscription half - an optimisation, since audio is subscribed by default, so attaching
-    // must never wait on it or fail with it
+    // the element already has the volume, so what is left to do here is the subscription - an
+    // optimisation, since audio is subscribed by default, so attaching must never wait on it
     track
       .setVolume(volume)
       // error, not warn: a warn is dropped once an app calls setLogLevel(ERROR)
--- a/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
+++ b/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
@@ -63,18 +63,12 @@ export class AudioSinkManager {
   }
 
   async setVolume(value: number) {
-    /**
-     * Validated here, not only per track: this.volume is recorded before the fan-out and the
-     * fan-out no longer rethrows, so a value the element setter refuses would be kept and then
-     * throw IndexSizeError on every later track add - before the element is wired up, leaving that
-     * peer with no audio element at all. `!(value >= 0 && value <= 100)` rather than the inverted
-     * range so NaN is rejected too.
-     */
+    // this.volume is recorded below and the fan-out no longer rethrows, so a value the element
+    // setter refuses would be kept and throw on every later track add. Not `< 0 || > 100`: NaN.
     if (!(value >= 0 && value <= 100)) {
       throw Error('Please pass a valid number between 0-100');
     }
-    // recorded before the fan-out: that is a round trip per track, and until it returns every
-    // track added meanwhile reads this - and a slow older call must not land after a newer one
+    // before the fan-out: tracks added while it is in flight read this
     this.volume = value;
     await this.store.updateAudioOutputVolume(value);
   }
@@ -143,15 +137,9 @@ export class AudioSinkManager {
     const audioEl = document.createElement('audio');
     audioEl.style.display = 'none';
     audioEl.id = track.trackId;
-    /**
-     * A fresh element starts at 1. Set the volume on the element itself, at the point it is
-     * created, rather than routing it through setVolume: that call also applies subscription state
-     * over the api data channel, and the element must carry the volume before play() whatever
-     * happens to the round trip.
-     *
-     * This runs again on decode-error recovery, so it has to prefer what was asked for on this
-     * track - otherwise the rebuild hands a peer the user muted individually back at full volume.
-     */
+    // a fresh element starts at 1, and it has to carry the volume before play() below. The track's
+    // own volume wins: this runs again on decode-error recovery, and the sink's would un-mute a
+    // peer the user muted individually.
     const volume = track.getRequestedVolume() ?? this.volume;
     audioEl.volume = volume / 100;
     audioEl.addEventListener('pause', this.handleAudioPaused);
@@ -182,13 +170,8 @@ export class AudioSinkManager {
     audioEl.srcObject = new MediaStream([track.nativeTrack]);
     callListener && this.listener?.onTrackUpdate(HMSTrackUpdate.TRACK_ADDED, track, peer);
     await this.handleAutoplayError(track);
-    /**
-     * The element already carries the volume; this is the subscription half, a round trip to the
-     * SFU. Audio is subscribed by default, so it is an optimisation - attaching and playing must
-     * never wait on it or fail with it, or a lost response leaves the track silent forever.
-     * Re-applies the same value the element got, so a rebuild does not resubscribe a peer the user
-     * muted individually.
-     */
+    // the subscription half - an optimisation, since audio is subscribed by default, so attaching
+    // must never wait on it or fail with it
     track
       .setVolume(volume)
       // error, not warn: a warn is dropped once an app calls setLogLevel(ERROR)
--- a/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
+++ b/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
@@ -63,8 +63,20 @@ export class AudioSinkManager {
   }
 
   async setVolume(value: number) {
-    await this.store.updateAudioOutputVolume(value);
+    /**
+     * Validated here, not only per track: this.volume is recorded before the fan-out and the
+     * fan-out no longer rethrows, so a value the element setter refuses would be kept and then
+     * throw IndexSizeError on every later track add - before the element is wired up, leaving that
+     * peer with no audio element at all. `!(value >= 0 && value <= 100)` rather than the inverted
+     * range so NaN is rejected too.
+     */
+    if (!(value >= 0 && value <= 100)) {
+      throw Error('Please pass a valid number between 0-100');
+    }
+    // recorded before the fan-out: that is a round trip per track, and until it returns every
+    // track added meanwhile reads this - and a slow older call must not land after a newer one
     this.volume = value;
+    await this.store.updateAudioOutputVolume(value);
   }
 
   /**
@@ -131,6 +143,17 @@ export class AudioSinkManager {
     const audioEl = document.createElement('audio');
     audioEl.style.display = 'none';
     audioEl.id = track.trackId;
+    /**
+     * A fresh element starts at 1. Set the volume on the element itself, at the point it is
+     * created, rather than routing it through setVolume: that call also applies subscription state
+     * over the api data channel, and the element must carry the volume before play() whatever
+     * happens to the round trip.
+     *
+     * This runs again on decode-error recovery, so it has to prefer what was asked for on this
+     * track - otherwise the rebuild hands a peer the user muted individually back at full volume.
+     */
+    const volume = track.getRequestedVolume() ?? this.volume;
+    audioEl.volume = volume / 100;
     audioEl.addEventListener('pause', this.handleAudioPaused);
 
     audioEl.onerror = async () => {
@@ -160,12 +183,14 @@ export class AudioSinkManager {
     callListener && this.listener?.onTrackUpdate(HMSTrackUpdate.TRACK_ADDED, track, peer);
     await this.handleAutoplayError(track);
     /**
-     * setVolume applies the subscription state over the API data channel, a round trip to the SFU.
-     * Audio is subscribed by default, so this is an optimisation - attaching and playing the track
-     * must never wait on it or fail with it, or a lost response leaves the track silent forever.
+     * The element already carries the volume; this is the subscription half, a round trip to the
+     * SFU. Audio is subscribed by default, so it is an optimisation - attaching and playing must
+     * never wait on it or fail with it, or a lost response leaves the track silent forever.
+     * Re-applies the same value the element got, so a rebuild does not resubscribe a peer the user
+     * muted individually.
      */
     track
-      .setVolume(this.volume)
+      .setVolume(volume)
       // error, not warn: a warn is dropped once an app calls setLogLevel(ERROR)
       .catch(error => HMSLogger.e(this.TAG, 'Could not apply audio subscription state', `${track}`, error));
   };
--- a/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
+++ b/packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
@@ -63,8 +63,9 @@ export class AudioSinkManager {
   }
 
   async setVolume(value: number) {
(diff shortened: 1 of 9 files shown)


packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -63,8 +63,10 @@ export class AudioSinkManager {
   }
 
   async setVolume(value: number) {
-    await this.store.updateAudioOutputVolume(value);
+    // recorded first: the fan-out below is a round trip per track, and until it returns every
+    // track added meanwhile reads this - and a slow older call must not land after a newer one
     this.volume = value;
```
cyril-k-031225: `shrink:` two lines of comment over one assignment. The *why* is one clause.

```suggestion
    // before the fan-out: tracks added while it is in flight read this
    this.volume = value;
```

cyril-k-031225 (commented):
Reviewed for over-engineering only — correctness is out of scope for this pass, and the fix itself looks right and is genuinely small: three real lines (`this.volume = value` moved up, `audioEl.volume` set at element creation, two statements swapped in `HMSAudioTrack.setVolume`).

What's cuttable is everything around those three lines:

- **~20 lines of comment prose** guarding them, most of it restating this PR description at the call site.
- **One test that is strictly subsumed** by the `play()`-time assertion next to it.
- **The fake-track factory written out twice** in the two new files — and a third time in the existing `remoteAudioSilenced.test.ts`.
- `Promise.allSettled` + `forEach` + a status branch where `.catch` inline does the same thing.

**net: −45 lines**, taking the PR from +241/−9 to +196/−9.

I applied all of it locally to check it holds:

- `41 suites / 294 tests` pass (was 295 — the one deleted test).
- The shrunk suites still fail **5 of 6** against pre-fix `27b6f9c5a`, so the regression guard survives the cuts.
- `tsc` and `eslint` clean.

Nothing here touches the behaviour being fixed. Happy to push the trimmed version to the branch if that's easier than applying the suggestions one by one.

<sub>Not included, since it predates this PR: `remoteAudioSilenced.test.ts` carries the third copy of the factory — another −12 if you want it folded into the same helper.</sub>

packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
```diff
@@ -131,6 +133,11 @@ export class AudioSinkManager {
     const audioEl = document.createElement('audio');
     audioEl.style.display = 'none';
     audioEl.id = track.trackId;
+    // A fresh element starts at 1. Set the volume on the element itself, at the point it is
+    // created, rather than routing it through setVolume: that call also applies subscription state
+    // over the api data channel, and the element must carry the volume before play() whatever
+    // happens to the round trip.
+    audioEl.volume = this.volume / 100;
```
cyril-k-031225: `shrink:` four lines of comment over one assignment, and it re-argues the `setVolume` ordering that the hunk below already explains.

```suggestion
    // a fresh element starts at 1, and it has to carry the volume before play() below
    audioEl.volume = this.volume / 100;
```

packages/hms-video-store/src/media/tracks/HMSAudioTrack.ts
```diff
@@ -31,11 +31,14 @@ export class HMSAudioTrack extends HMSTrack {
       throw Error('Please pass a valid number between 0-100');
     }
     this.requestedVolume = value;
-    // Don't subscribe to audio when volume is 0
-    await this.subscribeToAudio(value === 0 ? false : this.enabled);
+    // Local and immediate. Ordering this after the await left the element at the old volume for
+    // however long the SFU took to answer, and forever when it never did - and the subscribe call
+    // can now throw, which skipped it entirely.
```
cyril-k-031225: `shrink:` three lines of comment over a statement swap. The swap is the point; the incident history belongs in the commit message, which already has it.

```suggestion
    // local half first: the subscribe below can hang or throw
```

packages/hms-video-store/src/sdk/store/Store.ts
```diff
@@ -313,10 +313,24 @@ class Store {
...
-    for (const track of this.getAudioTracks()) {
-      await track.setVolume(value);
-    }
+    const results = await Promise.allSettled(this.getAudioTracks().map(track => track.setVolume(value)));
+    results.forEach(result => {
+      if (result.status === 'rejected') {
+        // error, not warn: a warn is dropped once an app calls setLogLevel(ERROR)
+        HMSLogger.e(this.TAG, 'could not apply audio subscription state for volume change', result.reason);
+      }
+    });
```
cyril-k-031225: `shrink:` a 9-line docblock restating the PR description, plus `allSettled` + `forEach` + a `status === 'rejected'` branch to do what `.catch` does inline. Same behaviour, half the lines, no result-shape unwrapping.

Also drops `// error, not warn: ...` — prose defending a log level.

```suggestion
  /**
   * One request to the SFU per track, in parallel. Logged not thrown, unlike updateAudioOutputDevice
   * below: setVolume already turned the elements down, so only the unsubscribe optimisation is lost.
   */
  async updateAudioOutputVolume(value: number) {
    await Promise.all(
      this.getAudioTracks().map(track =>
        track.setVolume(value).catch(error => HMSLogger.e(this.TAG, 'could not apply audio subscription state', error)),
      ),
    );
```

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    } as unknown as HMSSubscribeConnection;
+    const stream = new HMSRemoteStream({ id: 'stream-1' } as MediaStream, connection);
+    const nativeTrack = {
+      id: 'track-1',
+      kind: 'audio',
+      enabled: true,
+      addEventListener: jest.fn(),
+    } as unknown as MediaStreamTrack;
+    return new HMSRemoteAudioTrack(stream, nativeTrack, 'regular');
+  };
```
cyril-k-031225: `delete:` this fake-track factory is written out a second time in `Store.audioVolume.test.ts` (as `addTrack`) and a third time in the existing `remoteAudioSilenced.test.ts` — same three casts each time, different names.

One helper next to the existing `src/test/helpers/makeTransport.ts`:

```ts
// src/test/helpers/makeRemoteAudioTrack.ts
export type SubscribeBehaviour = 'answers' | 'hangs' | 'rejects';

const subscribeMock = {
  answers: () => jest.fn().mockResolvedValue({}),
  hangs: () => jest.fn(() => new Promise(() => undefined)),
  rejects: () => jest.fn().mockRejectedValue(new Error('No response from SFU')),
};

export const makeRemoteAudioTrack = (behaviour: SubscribeBehaviour, id = 'track-1') => { ... };
```

Call sites become `makeRemoteAudioTrack('hangs')` / `makeRemoteAudioTrack('answers')`. The boolean param goes too — `buildTrack(false)` doesn't say what `false` is.

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      audioSinkManager.cleanup();
+      window.MediaStream = realMediaStream;
+      jest.restoreAllMocks();
+    });
+
+    /**
+     * Attaching must not wait on the round trip - that was the silent-recording fix - but the
+     * element it attaches has to already carry the app's volume, or the peer is audible at full
+     * volume for as long as the SFU takes to answer, and forever if it never does.
+     */
```
cyril-k-031225: `delete:` two `/** */` blocks stacked on one `it`. Keep the second — it names the window that actually matters. This one repeats the file header.

```suggestion
```

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      // which would pass for any wrong-but-not-full value
+      expect(volumeAtPlay.length).toBeGreaterThan(0);
+      expect([...new Set(volumeAtPlay)]).toEqual([0]);
+    });
+
+    /**
+     * The element carries the volume because it is set where the element is created, not because
+     * setVolume got far enough to apply it - so a subscribe call that rejects outright, or a retry
+     * budget that runs out, cannot leave a peer audible.
+     */
```
cyril-k-031225: `shrink:` the test name already says it. Two lines of prose over a four-line test.

```suggestion
    /** The volume rides the element's creation, so a subscribe call that fails cannot undo it. */
```

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      const track = buildTrack(false);
+
+      await eventBus.audioTrackAdded.publish({ track, peer });
+      // publish does not await its subscribers, so the add is only part-way through the autoplay
+      // check when it returns - without this the assertions below never see the volume being set
+      await new Promise(resolve => setTimeout(resolve, 0));
+
+      expect(track.getAudioElement()?.srcObject).toBeTruthy();
+      expect(track.getAudioElement()?.volume).toBe(0);
+    });
```
cyril-k-031225: `delete:` strictly subsumed by `'is never audible at full volume, including at the moment play() is called'` above — same setup, same `hangs` connection, weaker assertion (settle-time `volume === 0` vs every observation at `play()`). The `srcObject` assertion is already carried by the reject test.

```suggestion
```

packages/hms-video-store/src/sdk/store/Store.audioVolume.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      id,
+      kind: 'audio',
+      enabled: true,
+      addEventListener: jest.fn(),
+    } as unknown as MediaStreamTrack;
+    const track = new HMSRemoteAudioTrack(stream, nativeTrack, 'regular');
+    track.setAudioElement(document.createElement('audio'));
+    store.addTrack(track);
+    return track;
+  };
```
cyril-k-031225: `delete:` second copy of the factory from `audioVolumeApplication.test.ts`. With the shared helper this is three lines:

```suggestion
  const addTrack = (id: string, behaviour: SubscribeBehaviour) => {
    const track = makeRemoteAudioTrack(behaviour, id);
    track.setAudioElement(document.createElement('audio'));
    store.addTrack(track);
    return track;
  };
```

packages/hms-video-store/src/sdk/store/Store.audioVolume.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  beforeEach(() => {
+    store = new Store();
+  });
+
+  it('applies the volume to the tracks after one whose request fails', async () => {
+    addTrack('first', true);
+    addTrack('unanswered', false);
+    const last = addTrack('last', true);
+
+    await expect(store.updateAudioOutputVolume(0)).resolves.toBeUndefined();
```
cyril-k-031225: `shrink:` `resolves.toBeUndefined()` asserts nothing `await` doesn't — a rejection fails the test either way, and the return type is `void`.

```suggestion
    await store.updateAudioOutputVolume(0);
```

packages/hms-video-store/src/sdk/store/Store.audioVolume.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      eventBus,
+    );
+    addTrack('unanswered', false);
+
+    try {
+      await expect(audioSinkManager.setVolume(0)).resolves.toBeUndefined();
+      expect(audioSinkManager.getVolume()).toBe(0);
+    } finally {
+      audioSinkManager.cleanup();
+    }
```
cyril-k-031225, line 52: `shrink:` `try/finally` guarding a `cleanup()` that is a no-op here — `init()` was never called, so there is nothing subscribed to tear down. Same `resolves.toBeUndefined()` note as above.

```suggestion
    await audioSinkManager.setVolume(0);

    expect(audioSinkManager.getVolume()).toBe(0);
```

packages/hms-video-store/src/audio-sink-manager/AudioSinkManager.ts
raviteja83: Applied verbatim.

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
raviteja83: Done — `src/test/helpers/makeRemoteAudioTrack.ts`, replacing all four copies (the two new files, `AudioSinkManager.test.ts`, and `remoteAudioSilenced.test.ts`). Outcome is a named union rather than a boolean, as you suggested; I used `'resolves' | 'rejects' | 'hangs'` since all three were needed.

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
raviteja83: Done — one block per test now.

packages/hms-video-store/src/media/tracks/audioVolumeApplication.test.ts
raviteja83: Done — prose dropped, and that test has since moved into `AudioSinkManager.test.ts` alongside the fixture it was duplicating.

packages/hms-video-store/src/sdk/store/Store.ts
raviteja83: Took the `allSettled` + `forEach` + status-branch out — you were right, that was three constructs for one job. But I went past the `.catch` too, and left it rejecting.

Two later passes landed on the same thing: the "only the unsubscribe optimisation is lost" premise holds when turning a peer *down*, not up. A resubscribe goes through `HMSRemoteStream.setAudio`, which records `this.audio` **before** the send — so a lost reply leaves the stream believing it is subscribed while the SFU sends nothing, and the `this.audio === enabled` guard makes every retry a no-op. Swallowing told the app the volume was applied while that peer stayed inaudible for the session, which is worse than the serial loop it replaced.

`Promise.all` with no catch keeps the concurrency fix (every track is asked, rather than stopping at the first failure) and keeps the signal — and is shorter than either version. The docblock is still 8 lines because that asymmetry with `updateAudioOutputDevice` is the non-obvious part; happy to cut it if you think the code carries it.

Leaving this one open for you.

cyril-k-031225 (dismissed):

cyril-k-031225 (approved):

PR #3745, opened 2026-08-26
fix(store): stop a stale sink attaching and keep the stats report unmutated

raviteja83:
**3 of 3.** Two files plus tests. Independent of the other two — no overlapping files, lands in any order.

### A stale sink attaches after the tile is gone
`VideoElementManager` stamps each intersection entry so a stale add loses to a newer one — but detaching the element (React unmounting a tile) or cleaning the manager up does not touch that stamp. An add suspended on its layer request resumed after the tile was gone, re-attached `srcObject` to an element nobody renders, and asked the SFU to keep streaming to it.

The check now also requires the element to still be registered, and `removeVideoElement` drops the stamp alongside the entry so a detach followed by a re-attach inside the same window cannot revive the stale add.

### A removed tile leaves the layer pinned
The layer asked for is the **max across live tiles**, so removing a tile can lower it — but `removeSink` asks for the unchanged `preferredLayer`, and against a stale one `shouldSendVideoLayer` sees the target as already current and sends nothing. A tile that pinned the layer to HIGH and was then unmounted left the SFU streaming HIGH for a tile that no longer exists.

`removeVideoElement` now recomputes from the tiles that remain *before* `removeSink` runs. The ordering is deliberate: deregister → recompute from survivors → `removeSink`. `sinkCount` is still non-zero during the recompute, so `setPreferredLayer` does not early-out on `hasSinks()`, and the layer it computes already excludes the dead tile.

### The stats report was being written to
`getActiveCandidatePairFromReport` hands back the browser's **own** `RTCStatsReport` candidate-pair entry, and the chain below it used `Object.assign(pair, …)` — which mutates the target, it does not clone. The report's entry was picking up `localCandidate`, `remoteCandidate` and `bitrate`/`outboundRtpBytesSent`, and the resolved candidates were handed out as the report's own objects for immer to deep-freeze in the store.

The same pattern was in `getRelevantStatsFromTrackReport` and `getTrackStats`, which read `inbound-rtp` and `remote-inbound-rtp` straight off the report and then assigned `remote`/`codec`/`qualityLimitationDurations`/`packetsLostRate` onto them — on every stats poll for every remote track. Both now build their own objects.

`handleResize` needed no change: `selectMaxLayer` iterates `videoElements`, not `entries`, so a resize for a detached element already contributes nothing. There is a test pinning that.

## Multi-sink behaviour is unchanged, and now covered
One track can be attached to several elements: `sinkCount` is a count on the track, `videoElements` is the manager's registry, and `intersectionSeq` stamps each element separately from one global counter — which is what keeps one element's churn from cancelling another's pending add. `selectMaxLayer` asks for a single layer, the max across live tiles.

Five new tests pin that, and they pass **both** on `main` and on this branch: the stream attaches to every element; the survivor stays attached when another is removed; the SFU is not told to stop while one tile still renders; NONE is sent only after the last element goes; and one element's intersection churn does not cancel another's add.

## Test plan
Same 12 tests run against `main`'s `VideoElementManager` and against this branch:

```
CURRENT PROD (origin/main)      4 failed, 8 passed
WITH THIS PR                    12 passed
```

The four that flip are the two teardown cases, the cleanup case, and the layer pinning. Every fix also has a mutation check. 39 suites / 301 tests pass on this branch alone; lint and tsc clean.

## Scope
Earlier revisions also excluded playlist and plugin tracks from the native mute/unmute listeners, and routed local-track `setEnabled` failures to the app. Both are dropped:

- **Playlist.** The premise was wrong. `PlaylistManager.pause()` only calls `element.pause()`; the captureStream track is then spec-muted, and the native `mute` listener is the *only* thing that publishes the pause to biz — `PlaylistManager` itself subscribes `handlePausePlaylist` to `localVideoEnabled`/`localAudioEnabled`. Removing the listener would have left remote peers rendering a frozen frame indefinitely. Playlist mute/unmute **are** paired, unlike screenshare.
- **Local-track errors.** `HMSLocalAudioTrack`/`HMSLocalVideoTrack` resolve to `undefined` inside `TrackManager` — a real import cycle, verified by probe — so `instanceof` throws before the rejection handler is attached and takes the process down. The store-based alternative misses the window between `store.addTrack` and the peer's `auxiliaryTracks.push`. Not worth a fragile check for a nice-to-have.

## Compatibility with 0.14.7
No public API change. The only behavioural changes are in the teardown paths above; multi-sink, attach, resize and the success paths are unchanged, with tests demonstrating that against `main`.

### DIFF
--- a/packages/hms-video-store/src/media/tracks/VideoElementManager.ts
+++ b/packages/hms-video-store/src/media/tracks/VideoElementManager.ts
@@ -234,9 +234,10 @@ export class VideoElementManager {
     if (maxLayer) {
       HMSLogger.d(this.TAG, `selecting max layer ${maxLayer} for the track`, `${this.track}`);
       /**
-       * Picking a layer is an optimisation over rendering the track, and the only callers are the
-       * resize and intersection handlers, which observers invoke without awaiting. Letting this
-       * reject would abort the sink attach and surface as an unhandled rejection.
+       * Picking a layer is an optimisation over rendering the track. The resize and intersection
+       * handlers call this without awaiting, and removeVideoElement awaits it only to order the
+       * recompute after removeSink - none of them can act on a failure. Letting it reject would
+       * abort the sink attach and surface as an unhandled rejection.
        */
       try {
         await this.track.setPreferredLayer(maxLayer);
--- a/packages/hms-video-store/src/media/tracks/VideoElementManager.ts
+++ b/packages/hms-video-store/src/media/tracks/VideoElementManager.ts
@@ -86,18 +86,22 @@ export class VideoElementManager {
   removeVideoElement(videoElement: HTMLVideoElement): void {
     this.videoElements.delete(videoElement);
     this.entries.delete(videoElement);
+    // dropped alongside the entry: without this, removing and immediately re-adding the same
+    // element leaves an in-flight add matching both the registry and its old stamp
     this.intersectionSeq.delete(videoElement);
     this.resizeObserver?.unobserve(videoElement);
     this.intersectionObserver?.unobserve(videoElement);
     /**
-     * removeSink first, and not behind an await: it nulls srcObject synchronously, and
-     * detachVideo/attachVideo run back to back when a tile swaps tracks - a deferred detach lands
-     * after the new stream is attached and nulls it.
+     * removeSink must run before the recompute, and must not be put behind an await: it nulls
+     * srcObject synchronously, and a tile swapping tracks detaches and re-attaches the same node
+     * without awaiting the detach - a deferred one lands after the new stream is attached and
+     * nulls it.
      *
-     * Then recompute. The layer asked for is the max across live tiles, so the tile being removed
-     * may be the one holding it up, and removeSink's own updateLayer asks for the unchanged
-     * `preferredLayer` - which shouldSendVideoLayer sees as already current and skips, leaving the
-     * SFU streaming for a tile that is gone.
+     * The recompute follows because the layer asked for is the max across live tiles, so the tile
+     * being removed may be the one holding it up, and removeSink's own updateLayer asks for the
+     * unchanged `preferredLayer` - which shouldSendVideoLayer sees as already current and skips,
+     * leaving the SFU streaming for a tile that is gone. The deletes above have to come first too,
+     * or the removed tile is still counted and the layer stays pinned.
      */
     // finally, not then: a removeSink that rejects is exactly when the layer is left pinned, so
     // skipping the recompute there abandons the case this exists for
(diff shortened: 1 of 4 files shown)


cyril-k-031225:
### Code review

Found 1 issue:

1. `removeVideoElement` now defers `removeSink` behind `await selectMaxLayer()`, but clears `videoElements`/`entries`/`intersectionSeq` synchronously — so the remove path has no staleness guard while the add path gained one. Any re-attach of the same `<video>` node inside that window (`useVideo.ts` does `detachVideo(prevTrackId, videoRef.current)` then `attachVideo(track.id, videoRef.current)` on the same node across a track swap, without awaiting the detach) completes first; the stale deferred `removeSink` then fires and, seeing only `srcObject !== null`, blanks the freshly attached tile and decrements the wrong track's `sinkCount`. Pre-PR this was `removeSink` as the first synchronous statement, so the window did not exist.

https://github.com/100mslive/web-sdks/blob/55ebc1419fd57db17e8b30f258eec7a68f3c3ffb/packages/hms-video-store/src/media/tracks/VideoElementManager.ts#L97-L102

`removeSink` has no equivalent of `shouldStillAddSink` — it checks the DOM, not the registry:

https://github.com/100mslive/web-sdks/blob/55ebc1419fd57db17e8b30f258eec7a68f3c3ffb/packages/hms-video-store/src/media/tracks/HMSVideoTrack.ts#L54-L60

Minor, not blocking: the JSDoc on `selectMaxLayer` still says "the only callers are the resize and intersection handlers, which observers invoke without awaiting" — `removeVideoElement` is now a third caller, and one that does depend on the result.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

raviteja83:
Thanks — the `removeSink` deferral was a real regression and it's fixed. You reviewed `55ebc1419`; the current tip does `removeSink` first and synchronously, then chains the recompute:

```js
Promise.resolve(this.track.removeSink(videoElement)).then(() => this.selectMaxLayer())
```

That keeps both properties: the detach stays synchronous, so the `useVideo` detach→attach-on-the-same-node window you described can't blank a freshly attached tile; and the layer still drops, because `removeSink`'s own `updateLayer` skips (target already current) and the recompute afterwards sends the lower layer.

Two tests pin it, both asserting with no `await` at all — the earlier ones all settled first, which is exactly why they missed this:

- `detaches the element synchronously, before any layer request`
- `does not null a stream attached to the same element after the removal`

Both fail if the order is swapped back.

Also fixed your minor point: the `selectMaxLayer` JSDoc now names `removeVideoElement` as the third caller and says it awaits only to order the recompute after `removeSink`.

cyril-k-031225 (approved):

PR #3751, opened 2026-09-07
fix(store): stop a mic track replacement from tearing down an audio plugin mid-add

cyril-k-031225:
# Description

Reported by Andor Health (Sentara rollout): Krisp appeared to be added repeatedly and the mic went silent, with `AddAlreadyInProgress` (7004) in the console.

Every native mic track replacement — device change, unmute after an empty track, and since 0.14.7 interruption recovery — runs `reprocessPlugins`: stop each plugin, rebuild the node graph, `init()` and `processAudioTrack()` each one again. That part is by design. The bug is that nothing stopped two of those from overlapping.

**Sequence on `main`** (matches the customer's log line for line, and reproduced with Prebuilt alone by switching mics twice within ~500ms):

1. Replacement A reprocesses: stops Krisp, re-inits it, and is inside `processAudioTrack` waiting on the filter node.
2. Replacement B reprocesses: `cleanup()` removes Krisp from the map and calls `stop()` on it — under A's in-flight add.
3. B's re-add is refused by A's `pluginAddInProgress` → `AddAlreadyInProgress` → `audioPluginFailed` → app error event.
4. A's add completes on a stopped plugin and publishes its output node. **The mic is silent, `getPlugins()` is empty, and the add resolved as a success.**

### What changed

**One queue, and one function that touches the graph.**

- `rebuildGraph()` is now the only thing that builds the node graph. The graph is a function of `pluginsMap`: stop whatever is wired up, then build `source → plugin₁ → … → destination` from the current list against the current native track. `addPlugin` appends to the map and calls it, `removePlugin` deletes and calls it, `reprocessPlugins` just calls it. There is no second, incremental add path to keep in sync, and — because `rebuildGraph` only ever runs inside a queued task — no public/`*Internal` pair of every method either.
- `addPlugin`, `removePlugin` and `reprocessPlugins` run their bodies through one promise queue, so a reprocess waits for an in-flight add and then rebuilds against the newest native track. Two overlapping replacements produce two sequential rebuilds, the last one winning on the current track. `pluginAddInProgress` is set *inside* the queued task, so it still means "an add is running" and never refuses an add that is only waiting its turn.
- **Teardown deliberately does not queue.** `HMSLocalAudioTrack.cleanup()` awaits `pluginsManager.cleanup()` *before* it stops the tracks, destroys the audio level monitor and removes its `visibilitychange` listener, and `RoleChangeManager` / `transport` await that whole method — so it must never be gated on plugin code, or a Krisp `init()` that does not settle wedges it. `cleanup()` sets `disposed` and tears down immediately; a rebuild still in flight checks `disposed` after each plugin it starts and stops what it started instead of publishing onto the dead graph. On `main` a leave during a pending add left the plugin initialised and registered on a torn-down graph, never stopped.
- **Per-plugin failures are contained instead of aborting or vanishing.** The room policy check and `checkSupport` both run per plugin on every rebuild, since a template change or a mic switch can turn either against a plugin mid-session. A plugin that cannot start is dropped, the rest of the graph is still built, and the first failure is thrown once the graph is up — which is what becomes `audioPluginFailed` for the app. On `main` the same situation either aborted the rebuild (destroying the other plugins' graph) or, in the reverse plugin order, warned and resolved as a success.
- A plugin is registered in `pluginsMap` only once it is actually wired up, so "registered" now means "running".

### Behaviour changes worth flagging

1. **`krisp.start` no longer fires on a reprocess**, only on a real app enable. Measured over two mic switches with Krisp on: `main` emits it **3×**, this branch **1×**. Intended — the old count conflated device changes with enables — but it is a step change in a customer-facing metric. (`clone()` → `handleSFUMigration` still re-publishes it through the public `addPlugin`, and no matching `krisp.stop` is published on teardown; both predate this PR.)
2. **A rebuild restarts the plugins it keeps.** Adding a second audio plugin now re-`init()`s the ones already running, where `main` appended to the chain without touching them. Krisp is the only audio plugin the SDK ships, so nothing in production hits this today; it is the deliberate cost of having one graph-building path instead of two.
3. **A plugin that fails to attach no longer resolves as a success.** `addPlugin` rejects and the native track is published, where `main` resolved and published an unfed destination node — the peer sent silence. This was listed as out of scope on the first pass; the rebuild's error handling closes it for free.

Otherwise unchanged: a second app `addPlugin` while one is actually in flight still throws `AddAlreadyInProgress`, and `removePlugin` semantics are the same.

## Test plan

`HMSAudioPluginsManager.test.ts`, 8 cases. Each of the first four was re-run with **its own** fix reverted and the others left in place, and fails only for its own:

| Test | Fails when |
|---|---|
| **reprocess vs in-flight add** — starts an add whose `processAudioTrack` is pending, swaps the native mic track and calls `reprocessPlugins` mid-flight; asserts the plugin is not stopped until the add completes, then that the reprocess rebuilt it once with a source node built from the new track | the queue is removed |
| **leave vs in-flight add** — starts an add whose `init` is pending, calls `cleanup()`; asserts cleanup resolves *without* the add having finished, then that the resumed add stops the plugin, never calls `processAudioTrack` and publishes nothing | `cleanup()` is put back on the queue (times out) |
| **remove vs in-flight add** — asserts mid-flight that `stop()` has not been called; the end-state assertion alone passes even when the remove runs early, so the mid-point is the test | `removePlugin` is taken off the queue |
| **adds queued behind a reprocess** — two app adds landing during a slow reprocess must both queue and both apply, not get `AddAlreadyInProgress` | `pluginAddInProgress` is set outside the queued task |

- **two-plugin reprocess with noise cancellation turned off**, run in both insertion orders: Krisp is dropped, the reprocess rejects in either order, and the other plugin keeps its rebuilt graph and stays published. On `main` the Krisp-first order destroyed the other plugin and the reverse order resolved silently.
- **`krisp.start` count** — one app enable plus two mic switches publishes it once, not three times (I4). Fails with the publish moved onto the rebuild path.
- **mid-rebuild registration** — a rebuild restarts a plugin without unregistering it, so `clone()`'s `pluginsMap.size > 0` gate never sees a transient zero. Asserted mid-flight inside the native-audio test.
- **a plugin that cannot attach** — `processAudioTrack` rejects: the add rejects, the plugin is unregistered and stopped, and the native track is published rather than a dead node.
- **single-plugin policy disablement then reprocess** — kept as a characterization test: it passes on `main` too, and guards the check being evaluated on the rebuild path (it fails if the check moves back into `addPlugin`).
- Full `hms-video-store` suite: 42 suites / 341 tests pass. Lint and `tsc` clean.
- Live: console script switching between two mics with a 150–550ms gap hit the bug on round 0 against the current release. Re-verify against this branch before release.

## Known, deliberately not in this PR

- The Krisp room-policy rejection is still a bare `Error`, so it reaches `onError` with `code` / `description` undefined. None of `MediaPluginErrors` (7001–7005) describes a room policy; it needs a code rather than a cast.
- The same overlap exists one level up: `HMSLocalAudioTrack.replaceTrackWith` stops the previous native track and acquires a new one with no serialization, so two concurrent `setSettings` calls can still interleave the *track* swap. Only the interruption path is guarded (`this.recovery`). The queue here makes the plugin half safe, not the track half.
- `utils/track-audio-level-monitor.ts` builds its analyser from the raw `nativeTrack` rather than `getTrackBeingSent()`, which is why this whole failure class is invisible in telemetry: the local meter bounces and `localAudioSilence` never fires while the published node emits silence.


raviteja83:
**Reviewed with the multi-agent PR toolkit — 1 blocker, 7 important.** The serialization itself is correct: rejections reach the original caller, FIFO holds, no unhandled rejections, no re-entrancy deadlock, and rewiring `validateAndThrow` → `cleanupInternal` correctly avoided a self-deadlock. Tests (a) and (b) are genuine regression guards — proved by reverting the fix, not assumed.

**The blocker is PR-introduced:** `cleanup()` now queues behind unbounded plugin work, so `RoleChangeManager.ts:100` → `transport/index.ts:628` can hang on Krisp's `init()`, and on leave the teardown tail after `HMSLocalAudioTrack.ts:428` never runs — mic indicator stays lit, `visibilitychange` listener never removed. Details below.

<details>
<summary><b>Critical (1)</b></summary>

### C1. `cleanup()` is gated on third-party plugin code — PR-introduced

`HMSAudioPluginsManager.ts:187` queues `cleanup()` behind whatever is in flight. `HMSLocalAudioTrack.cleanup():428` awaits it **first**, and `document.removeEventListener('visibilitychange', …)` is the **last** line of that method. There is no timeout in `serialize`, in `AudioPluginsAnalytics.initWithTime`, or anywhere else, so a Krisp `init()` that never settles wedges the queue permanently.

- `RoleChangeManager.ts:100` and `transport/index.ts:628` **await** it → role change into a non-publishing role blocks on plugin code.
- `sdk/index.ts:744` does **not** await → leave proceeds, but `super.cleanup()` has already stopped the native track while `processedTrack.stop()`, `tracksCreated.forEach(stop)`, `destroyAudioLevelMonitor()` and the listener removal never run. The still-attached listener can then reach `restoreCapture` → `getUserMedia`, re-opening the mic after the user left the room.

On `main`, `cleanup()` ran immediately and teardown was bounded by WebRTC, never by plugin code. Test 2 encodes the new behaviour as intended, so this needs a decision rather than a patch.

**Suggested shape:** a synchronous `disposed = true` set *outside* the queue, every `*Internal` bailing at its head, and `cleanupInternal()` running without waiting for in-flight plugin work. That yields the property test 2 actually wants — the in-flight add no-ops instead of resurrecting state — without the wait. The missing primitive here is pre-emption on the teardown path, not a stronger lock.

</details>

<details>
<summary><b>Important (7)</b></summary>

### I1. A two-plugin reprocess with NC disabled destroys every plugin

Measured against this branch:

```
before   = [ 'HMSKrispPlugin', 'Other' ]
reprocess => rejected: Krisp Noise Cancellation is not enabled for this room
after    = []          <- 'Other' silently gone, never re-initialised
```

`cleanupInternal()` empties the map, so `pluginsMap.size === 0` at `:102` is **always** true for the first plugin in the rebuild loop. It throws, the loop aborts, and `updateProcessedTrack()` at `:236` never runs. Same semantics on `main`, but this PR moves the check and adds a test covering only the single-plugin case, which hides it.

### I2. The reverse plugin order fails silently *and successfully*

With `[Other, Krisp]`, Krisp hits `size === 1` → `HMSLogger.w` + `return`. The reprocess **resolves**. NC is off; no `audioPluginFailed`, no `analytics.failure`, no `krispStop`.

Worse, OS-initiated device changes (`DeviceManager.ts:377`, and `:249` / `:538` / `:545`) call `audioTrack.setSettings(…, true)` directly, bypassing `HMSSDKActions.setAudioSettings:371` — the only caller of `syncRoomState('setAudioSettings'):376`. So the store keeps reporting `plugins: ['HMSKrispPlugin']` and the toggle stays on while Krisp is gone. That is the most common `reprocessPlugins` trigger in production.

I1 and I2 both come from inferring intent from `pluginsMap.size`. Pass an explicit `appInitiated` flag, `return` rather than `throw` on the reprocess path, and don't let one plugin's failure abort the rebuild loop.

### I3. `pluginAddInProgress` is set outside the queued task

`:82-88` — the flag now means "an add is *queued*", not "in progress". Behind a slow reprocess, unrelated adds get `AddAlreadyInProgress` (7004) — the error this PR exists to eliminate. The reject path at `:72-73` also calls `analytics.added()` then `analytics.failure()` → `clean(name)` on a name that may be a **live** plugin, destroying its analytics record so its `audioPluginStats` never emits.

Move it inside the serialized task, or drop it and let concurrent adds queue.

### I4. Undisclosed analytics change

Measured across two mic switches: `main` emits `krisp.start` **3×**, this branch **1×**. The new behaviour is right, but it is a step change in a customer-facing metric on rollout day and is not in the description. Two loose ends: `clone()` → `handleSFUMigration` still re-publishes `krispStart` through the public `addPlugin`, and `cleanupInternal` never publishes `krispStop` — so migrations emit a start with no matching stop.

### I5. No `disposed` guard, and `closeContext()` bypasses the queue

`closeContext()` at `:192` is public, unqueued, and nulls the one thing every queued task dereferences with `!`. An `addPlugin` queued **after** `cleanup()` reaches `addPluginInternal`, `initAudioNodes()` silently no-ops, and `this.audioContext!.sampleRate` at `:120` (or `:142`) throws a raw `TypeError` that reaches the app's `onError` instead of a typed `HMSException`. The queue makes this a guaranteed ordering rather than a race, so it is more reachable than on `main`, not less.

### I6. `validateAndThrow` is public and mutates the graph outside the queue

`:136` has no access modifier and calls `cleanupInternal()` at `:149` and `:157`, tearing down the whole node graph unqueued — exactly the interleaving this PR prevents. No external callers today (`HMSLocalAudioTrack.validatePlugin:391` routes to `validatePlugin`). Marking it `private` closes the hole and makes the doc block's invariant literally checkable.

### I7. `clone()` reads `pluginsMap` mid-reprocess

`HMSLocalAudioTrack.ts:100` gates plugin migration on `pluginsMap.size > 0`, which is transiently `0` during every rebuild (`cleanupInternal` empties it before the re-add loop). An SFU migration landing in that window migrates **zero** plugins and logs nothing.

</details>

<details>
<summary><b>Suggestions</b></summary>

- `this.queue.then(task, task)` at `:201` — the second argument is unreachable. `this.queue` is only ever `Promise.resolve()` or `run.catch(() => {})`, neither of which rejects. Use `.then(task)`.
- The empty catch at `:202-204` should log. All four callers consume `run` today, so nothing is lost — but that is an unenforced invariant, and a future non-awaiting caller would fail permanently silently, with the `.catch` also suppressing the unhandled-rejection warning.
- `:103` throws a bare `Error`. It reaches `onError` via an unchecked `as HMSException` cast at `HMSLocalAudioTrack.ts:327`, with `code` / `description` / `isTerminal` all `undefined`. Use `ErrorFactory.MediaPluginErrors.*`.
- **Doc comment `:44-45` is false in one direction.** `validateAndThrow` and `closeContext` are public and mutate the graph unqueued. Also worth stating the rule over `private` rather than the `*Internal` suffix — `initAudioNodes`, `updateProcessedTrack`, `processPlugin` and `connectToDestination` are equally forbidden from queueing but fall outside the rule as written.
- **Comment `:233` gives the wrong reason.** "with the in-progress guard this queue replaces" — the guard is retained, not replaced, and in an ordinary reprocess it would not fire anyway. The real reason not to call `addPlugin` there is a silent self-deadlock: `serialize` from inside a task chains onto a promise that only settles when that task settles. An `inTask` assertion would turn that hang into a loud failure.
- **Test gaps, in value order:** `removePlugin` is the one entry point moved onto the queue with no test — and an end-state test would pass on broken code, so it needs the mid-point `expect(plugin.stop).not.toHaveBeenCalled()`; a two-plugin reprocess (I1); a `krisp.start` count assertion (I4); test 2 has no assertion between `cleanup()` and `release()`, so despite its name it never asserts the waiting.
- Test 3 passes on `main`, so it is a characterization test rather than evidence of a fixed bug. Worth rewording in the description.

</details>

<details>
<summary><b>Out of scope here — worth separate tickets</b></summary>

- **`processPlugin` swallow publishes silence.** Already called out as deferred, and that is fair — but measured: `addPlugin` **resolves**, `getPlugins() === []`, `dest.connect` count `0`, and `setProcessedTrack({id:'processed'})` publishes an unfed destination node. It lives in the function this PR rewrote, and `reprocessInternal` now publishes the dead node twice. A refusal guard in `updateProcessedTrack` (throw if `outputTrack && !prevAudioNode`) closes this and two sibling paths in about six lines.
- **This whole bug class is invisible in telemetry.** `utils/track-audio-level-monitor.ts:38` builds its analyser from `new MediaStream([this.track.nativeTrack])` — the raw mic — while `getTrackBeingSent()` at `HMSLocalAudioTrack.ts:450` already returns `processedTrack || nativeTrack`. Every failure above leaves the native track loud while the published node emits silence, so `localAudioSilence` never fires, the local meter bounces normally, and the SFU sees valid Opus at the DTX floor. Pointing the monitor at `getTrackBeingSent()` would have caught the original report.

</details>

<details>
<summary><b>What was verified how</b></summary>

Tests were run against reverted variants rather than reasoned about:

| Variant | reprocess-vs-add | cleanup-vs-add | Krisp drop |
|---|---|---|---|
| This branch | pass | pass | pass |
| Full `main` revert | **FAIL** | **FAIL** | pass |
| `serialize()` neutralised | **FAIL** | **FAIL** | pass |
| Krisp check moved back to `addPlugin` | pass | pass | **FAIL** |

So (a) and (b) fail iff the queue is removed, and (c) fails iff the check leaves `addPluginInternal` — cleanly orthogonal, none passing for incidental reasons.

I1, I4 and the `processPlugin` dead-node result are measured outputs. C1, I2, I6, I7 and the telemetry point were traced against the code and re-checked by hand. `flush()` was checked as a barrier and is sound: the add needs 8 microtask ticks to reach `processAudioTrack`, and `stop` is still uncalled after 60 further ticks with the queue in place, so the mid-point assertion is meaningful rather than vacuous.

</details>

---

<sub>Automated multi-agent review (correctness, tests, error handling, comments). Findings are ranked, but judgement on scope is the author's.</sub>

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,242 @@ export class HMSAudioPluginsManager {
...
+      this.eventBus.analytics.publish(AnalyticsEventFactory.krispStop());
     }
-    this.pluginAddInProgress = true;
 
-    try {
-      await this.addPluginInternal(plugin);
-    } finally {
-      this.pluginAddInProgress = false;
+    await this.serialize(async () => {
+      if (!this.pluginsMap.has(name)) {
```
raviteja83, line 143: shouldn't this be the first check in this fn? we are tracking analytics even if the plugin was not there but we call remove on it?

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,242 @@ export class HMSAudioPluginsManager {
...
+    await this.updateProcessedTrack(undefined);
   }
 
-  // eslint-disable-next-line complexity
-  private async addPluginInternal(plugin: HMSAudioPlugin) {
-    const name = plugin.getName?.();
-    if (this.pluginsMap.get(name)) {
-      HMSLogger.w(this.TAG, `plugin - ${name} already added.`);
+  //Keeping it separate since we are initializing context only once
+  async closeContext() {
```
raviteja83, line 177: where is this used?

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
raviteja83: or move the analytics also inside after this?

raviteja83 (commented):
Reviewed across code / silent-failure / tests / comments. Every finding below was re-verified by an A/B probe against `origin/main` in a separate worktree, so the deltas are measured rather than reasoned.

**The core fix is real and reproduces.** A cleanup landing during a plugin re-`init()` ends `running=true, plugins=['A']` on main and `running=false, plugins=[]` here. `serialize()` is correct — the rejection reaches the caller via `run`, the chain can't wedge, and the derived `.catch` means no unhandled rejection. Removing main's two non-terminal `cleanup()` calls is also what makes the one-way `disposed` latch safe, and is worth keeping in the description. Unclaimed bonus: `clone()` issues `addPlugin` per plugin in a synchronous `forEach`, so before this every plugin after the first was lost to 7004 and logged as "Plugin add failed while migrating" — the queue fixes that.

**Blocking, in order:** the collateral-failure/wrong-error issue at :190, `cleanup()` being able to reject, and `publishGraph` still able to publish an unfed destination node. All inline.

### Test coverage
Revert matrix run one mechanism at a time — all five of the PR's claims hold, with a wider blast radius than stated (removing the queue fails 5 tests, not 1; re-queueing `cleanup()` fails 4, by timeout).

- **The `disposed` guards at :178 and :191 are both unpinned** — 40/40 pass with them deleted. Measured cost of removing them: post-teardown `setProcessedTrack` calls go 3 → 5.
- **The description's causal story for two tests is inverted.** Reverting the queue with only the *end-state* assertions deleted still passes for both the remove and the reprocess test. The mid-flight `expect(plugin.stop).not.toHaveBeenCalled()` is vacuous: deferred registration means `pluginsMap` is still empty, so the un-queued path takes the `!has(name)` branch and never calls `stop()` at all. What the queue actually prevents on the remove path is a **silently dropped `removePlugin`** — the plugin stays live and published after the user toggled it off. That is the better assertion, and a more interesting bug than the one claimed.
- **The interruption-recovery path — this PR's stated trigger — is never exercised.** Every new test drives `setSettings()`; nothing reaches `restoreCapture` / `reacquireStaleTrack`. The file already has the machinery for it.
- `getEmptyAudioTrack` (the permission-denied branch, newly reachable — see the `HMSLocalAudioTrack` comment) appears nowhere in the tests.
- `rejects.toThrow(/failed/)` passes for both a raw `processAudioTrack` error and the 7002-wrapped init error, so it can't see a change in what the app actually receives. `processAudioTrack` rejections propagate raw while init rejections arrive wrapped — no test asserts either code.
- No `beforeEach`/`clearAllMocks` in the manager suite; the module-level `audioContext` mock accumulates call history across all 18 cases. Harmless today, wrong for the next test that asserts "one source node per rebuild".

### Smaller things
- **Behaviour change #2 is understated.** `HMSAudioPlugin` is exported from the package's public `index.ts`, so customer-authored audio plugins are supported API surface — "Krisp is the only audio plugin the SDK ships" doesn't bound the blast radius of a rebuild restarting the plugins it keeps.
- **This failure class stays unmeasurable in prod after the fix ships.** `mediaPlugin.failed` only comes from `analytics.failure()`, which both `processAudioTrack` failures and the room-policy rejection skip; `unregister` publishes `mediaPlugin.stats` with `duration ≈ 0` instead, which is indistinguishable from a user toggling Krisp off a second after enabling it. Calling `analytics.failure` in `startPlugins`' catch would keep "how often does Krisp drop on a device switch" answerable.
- Two stale numbers in the description: "8 cases" → 13 `it` blocks / 18 cases; "42 suites / 341 tests" → 42 suites / **355** tests.
- The `eslint-disable-next-line complexity` main needed on `addPluginInternal` is genuinely no longer required — nice side effect.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -37,6 +37,20 @@ export class HMSAudioPluginsManager {
   private outputTrack?: MediaStreamTrack;
   private pluginAddInProgress = false;
   private room?: Room;
+  /**
+   * add, remove and reprocess all rebuild the node graph and stop/start the plugins, so they must
+   * not interleave: a reprocess (mic track replaced) landing while an add is still creating its
+   * filter node stops the plugin under that add, and the add then publishes a dead node. Those
+   * three are the only entry points that take this queue, and rebuildGraph - the one thing that
+   * touches the graph - only ever runs inside a task, so there is no queueing to duplicate.
```
raviteja83: "rebuildGraph - the one thing that touches the graph - only ever runs inside a task" isn't accurate: `cleanup()` at :133 calls `plugin.stop()`, `disconnectNodes()` and `updateProcessedTrack(undefined)` directly, and the very next docblock says it deliberately doesn't queue. The "only three entry points take this queue" half does check out.
Worth correcting, since a maintainer who trusts this line would add un-queued graph work and reintroduce the race.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
-          const errorMessage = 'Krisp Noise Cancellation is not enabled for this room';
-          if (this.pluginsMap.size === 0) {
-            throw Error(errorMessage);
-          } else {
-            HMSLogger.w(this.TAG, errorMessage);
-            return;
-          }
+    if (name === 'HMSKrispPlugin' && this.room?.isNoiseCancellationEnabled) {
+      // app enables only, a rebuild re-adding the plugin is not a new start
+      this.eventBus.analytics.publish(AnalyticsEventFactory.krispStart());
```
raviteja83: Verified A/B — two `addPlugin(krisp)` calls in the same tick emit **2** `krisp.start` here vs **1** on main, and the second no longer rejects 7004: `pluginAddInProgress` is assigned only at :92 *inside* the task, so the check at :73 still reads `false` at call time.
Publishing above the queue **and** above the `has(name)` check at :94 is what unpairs the metric (2 starts, 1 stop). Same ordering issue as the `krisp.stop` comment below, on the add side.
Suggest publishing inside the task, after :94 — that keeps the "one enable per app add" invariant the new test asserts. Note the description's "a second app addPlugin ... still throws AddAlreadyInProgress" no longer holds for same-tick calls, which is the realistic double-click case.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
+      plugin.stop();
+      this.unregister(name);
+      // the graph is rebuilt from what is left: there is no way to splice one plugin's nodes out
+      await this.rebuildGraph();
+    });
+  }
+
+  /**
+   * Deliberately not queued, see the disposed field. Track teardown awaits this before it stops the
+   * tracks and drops its listeners, so it can only be bounded by our own work, never by a plugin's.
```
raviteja83: Both halves of "awaits this before it stops the tracks ... can only be bounded by our own work, never by a plugin's" are off. `HMSLocalAudioTrack.cleanup()` runs `super.cleanup()` first, which is `HMSTrack.cleanup()` → `nativeTrack?.stop()`, so the capture track is already stopped before this is awaited.
And `plugin.stop()` at :136 *is* plugin work on the critical path — it's sync so it can't hang on a promise, but a throwing `stop()` aborts `cleanup()` mid-loop, skipping `analytics.cleanup()`, `disconnectNodes()` and the publish.
The real invariant is narrower and worth stating as such: nothing here *awaits* plugin code.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
+  async cleanup() {
+    this.disposed = true;
+    for (const [name, plugin] of Array.from(this.pluginsMap)) {
+      plugin.stop();
+      this.unregister(name);
     }
+    // Startup opens usage before registration and may still be pending at teardown.
+    this.analytics.cleanup();
+    this.disconnectNodes();
+    await this.updateProcessedTrack(undefined);
```
raviteja83: This can reject, and it's the last statement — so the rejection propagates into `HMSLocalAudioTrack.cleanup():429` and skips everything after it, including `destroyAudioLevelMonitor()` and `removeEventListener('visibilitychange', ...)`.
With that listener still attached to a torn-down track, the next tab-visible event walks `handleForegrounded → endInterruption → restoreCapture → reacquireStaleTrack → replaceTrackWith` and issues a fresh **`getUserMedia` after the user has left the room** — mic indicator on, no call in progress. Most callers (`sdk/index.ts:744`, `transport/index.ts:478`) don't even await `cleanup()`, so they'd never see the error either.
Teardown should be best-effort here — wrap in try/catch (`updateProcessedTrack` already logs at error). Belt and braces: make `HMSLocalAudioTrack.cleanup()` a try/finally so listener removal can't be skipped by anything.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
-    const name = plugin.getName?.();
-    if (this.pluginsMap.get(name)) {
-      HMSLogger.w(this.TAG, `plugin - ${name} already added.`);
+  //Keeping it separate since we are initializing context only once
+  async closeContext() {
+    this.audioContext = undefined;
+  }
+
+  async reprocessPlugins() {
+    await this.serialize(() => this.rebuildGraph());
```
raviteja83, line 182: main had `if (this.pluginsMap.size === 0 || !this.sourceNode) return;` here. Without it, `publishGraph` calls `updateProcessedTrack(undefined)` unconditionally on an empty map.
Measured on one device change with **no plugins at all**: 2 `sender.replaceTrack` calls vs 1 on main (`updateTrack` already did one at `HMSLocalAudioTrack.ts:259`), plus an Info-level "going back to the native track" on every device change, unmute and interruption recovery — for the majority of sessions that never add an audio plugin.
Suggest `if (this.pluginsMap.size === 0 && !this.outputTrack) { return; }`.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
-    try {
-      if (this.pluginsMap.size === 0) {
-        await this.initAudioNodes();
-      } else if (this.prevAudioNode) {
-        // Previous node will be connected to destination. Disconnect that
-        this.prevAudioNode.disconnect();
+    let failure: Error | undefined;
+    if (plugins.length > 0) {
+      this.initAudioNodes();
+      failure = await this.startPlugins(plugins);
```
raviteja83: `startPlugins` returns the *first* failure from anywhere in the chain, and :199 throws it — so an add is rejected with an unrelated plugin's error.
Verified: A running, A's forced restart fails, then `addPlugin(B)` → rejects with `"A restart failed"`, `getPlugins()` is `["PluginB"]`, and B is running and published. The caller that asked for B gets a failure about A, B is actually live, and A was dropped with no event at all. An app that flips its toggle off on rejection now desyncs from the real graph. `removePlugin` has the mirror problem when a surviving plugin fails to restart.
Suggest keying failures by plugin name: reject the add/remove with that plugin's own failure, and report collateral ones through `eventBus.audioPluginFailed`.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
+      if (this.disposed) {
```
raviteja83, line 249: This branch discards `failure`, so `addPlugin`/`reprocessPlugins` resolve as **success** for a plugin that never started — and prebuilt then persists the preference off that. One line: `if (this.disposed) { if (failure) throw failure; return; }`.
Separately: deleting this guard **and** the one at :178 leaves 40/40 tests passing, so half the `disposed` design is unpinned. A test that snapshots `setProcessedTrack.mock.calls.length` after `cleanup()` with two rebuilds queued, drains, then asserts unchanged, pins both — measured 3 → 3 with the guards, 3 → 5 without.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
+      await this.updateProcessedTrack(undefined);
+      return;
     }
-    await this.hmsTrack.setProcessedTrack(undefined);
-    //disconnect nodes, stop track
-    this.sourceNode?.disconnect();
-    this.prevAudioNode?.disconnect();
-    this.outputTrack?.stop();
+    this.connectToDestination();
+    await this.updateProcessedTrack(this.outputTrack);
```
raviteja83: `connectToDestination()` swallows a failed `connect()` at error level and this line publishes `outputTrack` regardless — an unfed destination node, i.e. exactly the "publishes silence and resolves as success" mode this PR exists to fix, reached by a different route. Worth closing while you're here, since nothing client-side can detect it: `TrackAudioLevelMonitor` analyses `track.nativeTrack`, so the local meter and `localAudioSilence` both read the healthy mic.
`processAudioTrack`'s return is also never validated at :282, so a plugin resolving `undefined` leaves `prevAudioNode` falsy — the guard then silently connects nothing and doesn't even log.
Suggest throwing when `!prevAudioNode || !destinationNode`, letting `connect()` throw, and reusing the `size === 0` path to fall back to the mic.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
-    this.audioContext = undefined;
+  /**
+   * Everything that has to hold every time a plugin is wired into the graph, on a fresh add and on
+   * every rebuild alike: the room policy and the device support are both re-checked, because a
+   * template change or a mic switch can turn either of them against a plugin mid-session.
+   */
+  private async startPlugin(plugin: HMSAudioPlugin) {
+    const name = plugin.getName?.();
+    if (name === 'HMSKrispPlugin' && !this.room?.isNoiseCancellationEnabled) {
+      // ponytail: no MediaPluginErrors code covers a room policy, so this stays untyped for now
```
raviteja83: `// ponytail:` looks like a stray marker — it's the only occurrence anywhere in `packages/`, and the convention in this directory is `// TODO:` (`HMSVideoPluginsManager.ts:305`, `VideoPluginsAnalytics.ts:95`). The prose after it is worth keeping.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -67,152 +81,244 @@ export class HMSAudioPluginsManager {
...
+      throw Error('Krisp Noise Cancellation is not enabled for this room');
```
raviteja83: Confirmed this reaches **users**, not just `onError`, which I think moves it out of "deliberately not in this PR": bare `Error` → `adapter.ts:207` copies `description: undefined` → `isTerminal`/`code`/`action` all undefined, so it falls through every guard in `ErrorNotifications.tsx` to :51, `title: Error: ${message} - ${description}`.
The Prebuilt toast reads literally **"Error: Krisp Noise Cancellation is not enabled for this room - undefined"** (gated on `subscribedNotifications.ERROR`, on by default).
It also blocks the failure telemetry above, since `analytics.failure()` calls `error.toAnalyticsProperties()`, which a bare `Error` doesn't have. A real `MediaPluginErrors` code fixes both.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -259,6 +259,12 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
     await this.replaceSenderTrack();
     const isLevelMonitored = Boolean(this.audioLevelMonitor);
     isLevelMonitored && this.initAudioLevelMonitor();
+    // Recovery tracks need the same graph rebuild as successful device selections.
+    try {
+      await this.pluginsManager.reprocessPlugins();
```
raviteja83, line 297: Right call for two of the three `updateTrack` callers — :294 (success) and :320, where it's a genuine fix: main's `throw` skipped the old end-of-`replaceTrackWith` reprocess, leaving a live fallback mic feeding a dead graph.
The third is wrong. :305 is the `CANT_ACCESS_CAPTURE_DEVICE` / `SYSTEM_DENIED_PERMISSION` branch, which installs `LocalTrackManager.getEmptyAudioTrack()` — a deliberately `enabled=false` oscillator. Verified this now runs a full Krisp `init()` + `processAudioTrack` against it (source stream built from the empty track, processed node published) *before* the device error is rethrown, so it also delays the app's error handling by a model load.
Suggest guarding the reprocess with `isEmptyTrack(track)`.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +79,387 @@ export class HMSAudioPluginsManager {
...
+        await this.updateProcessedTrack(undefined).catch(() => {});
       }
+      return;
     }
+    if (this.disposed) {
+      return;
+    }
+    HMSLogger.i(this.TAG, 'no plugin output to publish, going back to the native track');
+    this.disconnectNodes();
+    await this.updateProcessedTrack(undefined);
```
raviteja83, line 369: When `connectToDestination()` fails we publish the native track, but the plugins stay in `pluginsMap` **and stay started** — `disconnectNodes()` only unwires the Web Audio nodes, nobody calls `plugin.stop()`. So `getPlugins()` keeps reporting Krisp as on, no `audioPluginFailed` goes out, and the app has no way to see it has been bypassed.

The trigger is narrow: `!prevAudioNode` with `size > 0` is unreachable (every successful `startPlugin` sets it, and dropped ones get unregistered), so it needs a plugin returning a node from a foreign context, or `connect()` throwing. And it self-heals on the next reprocess, since `outputTrack` is undefined and the map is non-empty. But until then the SDK's own state is wrong.

Suggest stopping + unregistering the plugins on this path and routing a failure through `reportFailures`, the same as a plugin that could not start.

raviteja83 (commented):
Reviewed the plugin manager for race conditions around the unqueued `cleanup()` path. The teardown interleavings themselves look well covered — cleanup during `stopGraph`'s republish, during `plugin.init()`, during `processAudioTrack`, during the final `setProcessedTrack`, and for tasks still sitting in the queue all unwind correctly, and analytics usage closes exactly once (`removed`/`failure` both guard on `pluginAdded[name]` and `clean()` deletes it, so the double calls are no-ops). `disposed` is genuinely needed: queueing cleanup would block teardown for up to ~2N x `AUDIO_PLUGIN_CALL_TIMEOUT` with the mic live.

What I'd still change is inline — two of them are one-liners.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +79,387 @@ export class HMSAudioPluginsManager {
...
+      try {
+        this.throwIfDisposedForAdd(plugin);
+        if (this.pluginsMap.has(name)) {
+          HMSLogger.w(this.TAG, `plugin - ${name} already added.`);
+          return;
+        }
+        if (name === 'HMSKrispPlugin' && this.room?.isNoiseCancellationEnabled) {
+          // one per app enable: neither a rebuild re-adding the plugin nor an add of a plugin that
+          // is already running is a new start, and krisp.stop is only published for a real removal
+          this.eventBus.analytics.publish(AnalyticsEventFactory.krispStart());
```
raviteja83: `rebuildGraph(plugin)` below can still throw — init timeout, unsupported device, room policy — and by then `krispStart` is already out with no `krispStop` to match it: `cleanup()` does not publish one by design, and a later `removePlugin` early-returns because the failed plugin was never registered. Every failed Krisp enable becomes a permanent unmatched start.

Moving the publish to after `await this.rebuildGraph(plugin)` resolves keeps the "one per app enable" property and fixes the asymmetry.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +79,387 @@ export class HMSAudioPluginsManager {
...
-          if (this.pluginsMap.size === 0) {
-            throw Error(errorMessage);
-          } else {
-            HMSLogger.w(this.TAG, errorMessage);
-            return;
-          }
+    await this.serialize(async () => {
+      // set inside the task: the flag is the app facing "an add is running" guard, it must not also
+      // cover the time this task spends queued behind a reprocess, or an unrelated add gets 7004
+      this.pluginAddInProgress = true;
```
raviteja83, line 106: The flag is read outside the queue (L77) and set here, inside the task — so two `addPlugin` calls in the same tick both see `false` and both get queued. The queue keeps that safe, but it makes the `AddAlreadyInProgress` contract non-deterministic: a second add succeeds if it arrives while the first is still queued behind a reprocess, and gets 7004 if it arrives a tick later.

If the 7004 is meant to be deterministic, this needs a counter incremented at call time. If best-effort is the intent, worth saying so in the comment — as written it reads like mutual exclusion.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -256,9 +256,32 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
+     * The empty track installed when capture is denied is a deliberately silent oscillator. Running
+     * the plugins against it only delays the device error the caller is about to get by a model
+     * load, and the graph that is published right now reads the mic that has just gone away, so
+     * publish the placeholder itself instead of a node fed by a stopped track.
+     */
+    const isSilentPlaceholder = isEmptyTrack(track);
+    if (isSilentPlaceholder) {
+      // drop the running graph (Krisp keeps processing a stopped mic otherwise) but do not
+      // rebuild against the oscillator — plugins stay registered for the next real capture
+      await this.pluginsManager.releaseGraph();
```
raviteja83: `releaseGraph()` can reject: `stopGraph` -> `updateProcessedTrack` rethrows on a `replaceSenderTrack` failure. On this path that rejection propagates out of `updateTrack` before `replaceTrackWith`'s `throw error` at L323, so `SYSTEM_DENIED_PERMISSION` / `CANT_ACCESS_CAPTURE_DEVICE` gets replaced by a plugin error — which is exactly the error this branch exists to surface.

`reprocessPlugins` at L281 is already wrapped for this reason; this one is not.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +79,387 @@ export class HMSAudioPluginsManager {
...
+    if (this.pluginsMap.size > 0 && this.connectToDestination()) {
+      await this.updateProcessedTrack(this.outputTrack);
+      if (this.disposed) {
+        // cleanup already published native; this publish may have landed after it
+        await this.updateProcessedTrack(undefined).catch(() => {});
       }
+      return;
     }
+    if (this.disposed) {
+      return;
```
raviteja83: This check cannot fire. `publishGraph` has one caller, and both paths into L226 already check `disposed` with no `await` in between (L204 when the map is empty, L218 otherwise) — and `pluginsMap.size` and `connectToDestination()` are both synchronous, so there is no yield point for `cleanup()` to land in between L276 and here.

Same applies to L276 itself. Only the post-`updateProcessedTrack` check at L282 is load-bearing. Worth collapsing to that one, or marking the others as deliberately defensive so the next reader does not have to re-derive it.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +79,387 @@ export class HMSAudioPluginsManager {
...
+      plugins.push(added);
+    }
+
+    let failures = new Map<string, HMSException>();
+    if (plugins.length > 0) {
+      this.initAudioNodes();
+      failures = await this.startPlugins(plugins);
+      if (this.disposed) {
+        // startPlugins unwound what it had started, a cleanup published the native track already
+        this.reportFailures(failures, added);
```
raviteja83, line 251: Minor: on the disposed path these collateral failures still go out on the event bus after `cleanup()` has run, so the app can receive an `audioPluginFailed` for a track that no longer exists. Probably harmless, but it is the one place the disposed unwind still publishes.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -37,6 +39,22 @@ export class HMSAudioPluginsManager {
   private outputTrack?: MediaStreamTrack;
   private pluginAddInProgress = false;
   private room?: Room;
+  /**
+   * add, remove and reprocess all rebuild the node graph and stop/start the plugins, so they must
+   * not interleave: a reprocess (mic track replaced) landing while an add is still creating its
+   * filter node stops the plugin under that add, and the add then publishes a dead node. Those
+   * three are the only entry points that take this queue, and rebuildGraph - the only thing that
+   * builds the graph - only ever runs inside one of their tasks, so there is no queueing to
+   * duplicate. cleanup is the one exception, it tears the graph down unqueued, see disposed.
```
raviteja83, line 57: Structural note on this comment: the queue serializes the three graph builds against each other, but not against the native track swap itself. `this.nativeTrack = track` in `HMSLocalAudioTrack.updateTrack` is outside it, so a rebuild already in flight can snapshot the old, stopped track in `initAudioNodes` and publish silence.

That is correct as written only because `reprocessPlugins()` is always queued behind the swap, which bounds the silence to the in-flight add's remaining init (up to `AUDIO_PLUGIN_CALL_TIMEOUT`). Worth stating that dependency here — the comment covers the interleaving between add/remove/reprocess, but not this one.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Fixed in d7ddd2a4e, as suggested. `publishGraph` now stops, unregisters and reports the plugins on that path, so the add rejects instead of resolving onto a bypassed graph.

Reused `ProcessingFailed` (7003) rather than adding a code, and routed it through the `failures` map so it lands in `reportFailures` — the add gets its own failure, collateral ones go out as `audioPluginFailed`, same as a plugin that could not start. Ordered `analytics.failure` before `unregister` so the usage interval is still live, as on the start path.

Test: `HMSAudioPluginsManager.test.ts:567`, forcing the mismatch with a node from a foreign context.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Fixed in d7ddd2a4e — the publish moved below `await this.rebuildGraph(plugin)`, so a failed enable emits nothing.

Worth noting for whoever reads the analytics: this was never a billing problem, since Krisp minutes dedupe by peer rather than by event count. It was the start/stop pairing that was wrong, which is what makes "how often does a Krisp enable fail" answerable at all.

Test: `HMSAudioPluginsManager.test.ts:585`.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Best effort is the intent, so I documented it rather than adding a counter — d7ddd2a4e moves the explanation onto the field, where both the read and the write point.

A call-time counter would actually be a regression: the queued-adds test at `HMSAudioPluginsManager.test.ts:420` asserts that two adds arriving while a reprocess is still running both succeed. Refusing the second one there is the worse failure — it is an unrelated plugin being told an add is in progress when nothing is running yet. So 7004 only refuses an add that arrives once an earlier one is genuinely running, and the queue, not the flag, is what keeps them from interleaving.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
cyril-k-031225: Fixed in d7ddd2a4e, and widened slightly — `setProcessedTrack(undefined)` on the next line reaches `replaceSenderTrack` too, so both calls are inside the `try`.

The wrap alone was not enough though. `stopGraph` awaits the native publish *before* the loop that stops the plugins, so the rejection was abandoning the teardown: plugins left running, nodes left wired. Swallowing it would have surfaced the device error while leaving Krisp on the mic that had just been stopped — exactly what this branch exists to prevent. So `stopGraph` is now `try/finally`.

Test at `HMSLocalAudioTrack.test.ts:569` asserts both halves: `setSettings` still rejects with the device error, and `plugin.stop` still ran.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Agreed, and confirmed while restructuring for the `connectToDestination` fix above — `pluginsMap.size` and `connectToDestination()` are both sync, so nothing could land in between.

d7ddd2a4e drops this one (the `size > 0` branch is nested now, so it falls out naturally) and labels the entry check as deliberately defensive against a future second caller. The post-`updateProcessedTrack` check is the load-bearing one and stays.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Fixed in d7ddd2a4e — one-line `else if (!this.disposed)` in `reportFailures`, with a note that the track it would be reported against is gone. Analytics still has the failure from `startPlugins`, so nothing is lost for counting.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
cyril-k-031225: Good catch, added in d7ddd2a4e. The docblock now says the queue does not serialize against the swap itself, that a rebuild in flight can have snapshotted the stopped track in `initAudioNodes`, and that what bounds the silence is `reprocessPlugins` being queued behind the swap — up to `AUDIO_PLUGIN_CALL_TIMEOUT`. Ends with "keep that reprocess queued rather than immediate", since making it immediate is the plausible-looking change that would break it.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -256,9 +256,39 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
     // change nativeTrack so plugin can start its work
     this.nativeTrack = track;
-    await this.replaceSenderTrack();
+    /**
+     * The empty track installed when capture is denied is a deliberately silent oscillator. Running
+     * the plugins against it only delays the device error the caller is about to get by a model
+     * load, and the graph that is published right now reads the mic that has just gone away, so
+     * publish the placeholder itself instead of a node fed by a stopped track.
+     */
+    const isSilentPlaceholder = isEmptyTrack(track);
```
raviteja83, line 265: `isEmptyTrack` is an OR — `'canvas' in track || label === 'MediaStreamAudioDestinationNode' || label === ''` — so the third disjunct fires on its own for a real, capturing mic that reports no label.

Measured: a device switch to a live track that is normal in every way except `label: ''`.

```
plugin.init calls: 1   plugin.stop calls: 1
getPlugins(): ["HMSKrispPlugin"]   sender: native track
```

Krisp is stopped, the `return` at :283 skips `reprocessPlugins`, and it never comes back — while `getPlugins()` still reports it on and no `audioPluginFailed` is published. That is the bypassed-but-reported-healthy mode `publishGraph` was rewritten to close, reached from the other side, and it repeats on every subsequent switch. On `main` this track went through `replaceSenderTrack` + reprocess and kept working.

Tightening the predicate to the destination-node label alone does not work either: per that function's own comment, Firefox reports `''` for the destination track, so it would miss the real placeholder there and run Krisp against the oscillator — the failure this branch was added to prevent.

The track cannot tell the two apart, but the producer can. `LocalTrackManager.getEmptyAudioTrack()` is the only thing that makes one:

```ts
const emptyAudioTracks = new WeakSet<MediaStreamTrack>();
// in getEmptyAudioTrack, before returning
emptyAudioTracks.add(emptyTrack);
export const isEmptyAudioPlaceholder = (t: MediaStreamTrack) => emptyAudioTracks.has(t);
```

Exact and browser-independent, and it splits the two predicates apart: `isEmptyTrack` stays the permissive "might not be capturing" check that `shouldReacquireTrack` wants, where a false positive only costs a re-acquire, and this branch gets the strict one, where a false positive is a silent Krisp outage with no telemetry.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -256,9 +256,39 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
+    if (isSilentPlaceholder) {
+      // drop the running graph (Krisp keeps processing a stopped mic otherwise) but do not
+      // rebuild against the oscillator — plugins stay registered for the next real capture.
+      // Best effort: the only caller of this branch is about to throw the device error the app
+      // needs to see, and both calls below reach sender.replaceTrack, so a failed swap here must
+      // not take its place. updateProcessedTrack already logs the underlying failure at error.
+      try {
+        await this.pluginsManager.releaseGraph();
+        await this.setProcessedTrack(undefined);
```
raviteja83: These two share a `try`, and `setProcessedTrack` clears `this.processedTrack` *before* it swaps the sender. So when `releaseGraph()` rejects — `stopGraph` -> `updateProcessedTrack(undefined)` -> `replaceSenderTrack` — the field is already `undefined`, `disconnectNodes()` has stopped `outputTrack`, and the line below never runs.

Measured on the existing "still surfaces the device error" test, which only asserts `plugin.stop`:

```
sender.track.id: "processed"   readyState: "ended"
getTrackBeingSent(): "empty"   replaceTrack calls: 2
```

The peer is left on an **ended** track while the SDK believes it is publishing the placeholder, and the retry that would have succeeded (the mock rejects once) is skipped because the catch swallowed the first failure. Give the second call its own catch, so a failed teardown swap cannot take the sender reset with it.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
```diff
@@ -256,9 +256,39 @@ export class HMSLocalAudioTrack extends HMSAudioTrack {
...
     const isLevelMonitored = Boolean(this.audioLevelMonitor);
     isLevelMonitored && this.initAudioLevelMonitor();
+    if (isSilentPlaceholder) {
+      return;
+    }
+    // Recovery tracks need the same graph rebuild as successful device selections.
+    try {
+      await this.pluginsManager.reprocessPlugins();
+    } catch (e) {
+      this.eventBus.audioPluginFailed.publish(e as HMSException);
```
raviteja83, line 299: `startPlugins` normalises everything through `toPluginError`, but `updateProcessedTrack` rethrows raw — a `DOMException` from `sender.replaceTrack` propagates out of `stopGraph`/`publishGraph`, through `rebuildGraph` and the queued task, and lands here. The cast is then a lie: the app reads `code` and `description` as `undefined`, and Prebuilt renders `Error: <message> - undefined`, the same toast the room-policy error produced before you gave it 7006.

The cast moved into the diff along with the reprocess, and `toPluginError` already exists for exactly this — worth exposing it or normalising here.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +95,439 @@ export class HMSAudioPluginsManager {
...
+        // which no client side check can see: the level monitor reads the mic. Drop the plugin.
+        throw Error(`plugin ${name} returned no audio node`);
+      }
+      // the previous plugin was the end of the chain, extend it with this one
+      this.prevAudioNode?.connect(currentNode);
+      this.prevAudioNode = currentNode;
     } catch (err) {
-      HMSLogger.e(this.TAG, 'failed to add plugin', err);
+      // This startup may own resources even if an earlier instance was already stopped.
+      plugin.stop();
```
raviteja83: Every other stop in this PR goes through `stopPlugin()`, which catches. This one does not, so a `stop()` that throws on a half-initialised plugin replaces the failure being reported: `throw err` never runs, `startPlugins`' catch sees the `stop()` error instead, `toPluginError` wraps it as 7003, and `analytics.failure` records the wrong cause — the app never learns init failed (7002). The same masking applies to the `returned no audio node` throw above.

`this.stopPlugin(name, plugin, 'after a failed start')`.

packages/hms-video-store/src/plugins/audio/HMSAudioPluginsManager.ts
```diff
@@ -61,158 +95,439 @@ export class HMSAudioPluginsManager {
...
+        // toAnalyticsProperties off it, and an untyped error would throw out of this catch and cost
+        // the healthy plugins the graph this rebuild has already dismantled
+        const failure = this.toPluginError(err);
+        HMSLogger.e(this.TAG, `failed to start plugin ${name}, dropping it`, failure);
+        // before unregister ends the usage interval: otherwise this is reported as a plugin the user
+        // turned off right after enabling it, and "how often does Krisp drop" stays unanswerable
+        this.analytics.failure(name, failure);
+        this.unregister(name);
+        failures.set(name, failure);
+        continue;
```
raviteja83, line 401: `continue` skips the `disposed` check below. If a teardown lands during A's failing `init()`, the loop goes straight into `startPlugin(B)` and runs `analytics.added`, `checkRoomPolicy`, `validateAndThrow` and a full `await init()` — up to `AUDIO_PLUGIN_CALL_TIMEOUT` — against a torn-down track, before the post-call check unwinds it.

The non-failing path is fine, for what it is worth: there is no yield point between `pluginsMap.set` and the next `startPlugin`, so nothing can land there. A `disposed` check at the top of the loop body covers both.

raviteja83 (commented):
> **Correction:** the `isEmptyTrack` finding at `HMSLocalAudioTrack.ts:265` is **retracted** — 24h of `client.device.list.success` shows 0 blank audio-input labels in 269,308 permissioned listings, so the input it assumed does not occur. Details in that thread. **Four** blocking items, not five.

Re-review of `ac27f327d` (`d7ddd2a4e` plus a merge of main that does not touch `hms-video-store`). Baseline on this head: **42 suites / 373 tests pass, `tsc` and `eslint` clean.** Every round-1 and round-2 item is genuinely closed — `reportFailures`, the `cleanup()` catch, `publishGraph`'s bypass drop, the `disposed` snapshot test, the interruption-recovery test, `clearAllMocks`.

Five things left, all inline. Two are measured with a probe in a worktree at this head; three are read off the code.

### Checked and not raising

Each of these behaves identically on `origin/main`, so none is introduced here:

- `removePlugin` stopping the plugin before `stopGraph` republishes native — main's `removePluginInternal` had the same ordering.
- Two `setProcessedTrack(undefined)` when the last plugin goes — main did the same via `cleanup()` plus the explicit call in `removePlugin`.
- `initAudioNodes`' `!this.destinationNode` guard — same guard on main, though it is genuinely dead now that `stopGraph`'s `finally` always nulls it. Worth dropping for clarity, not for correctness.

Two I want to argue *against* changing:

- **The reprocess awaiting the queue is not a regression.** Main's `reprocessPlugins` also awaited `init` + `processAudioTrack` per plugin, with no timeout at all; this PR adds the 30s cap. The only case main returned faster is 7004 while an add was in flight — and that is the silent mic this PR exists to fix.
- **Leaving the 7004 path without analytics is right.** Publishing `mediaPlugin.failed` there would now close the *running* plugin's usage interval, because this PR makes that event interval-closing (`PluginUsageTracker.ts:38`). Main's version was worse again: it called `analytics.added` first and clobbered the live record. Worth saying in the comment at :98 that the tracker change is what makes this load-bearing.

### Non-blocking

- `removePlugin` now early-returns before publishing `krisp.stop` when the plugin is already unregistered, so a policy drop or a failed restart leaves an unmatched `krisp.start` — main's was unconditional. Pairing only, but it is the metric the publish reordering was for.
- `HMSLocalAudioTrack.cleanup()` puts `super.cleanup()` inside the `try`, so a throw there skips `pluginsManager.cleanup()` and `disposed` is never set. `HMSAudioTrack.cleanup()` is `nativeTrack?.stop()` plus an `audioElement` block that is null on a local track, so this is theoretical — but the `finally` exists precisely so nobody has to re-derive that.
- The description is stale for what is about to become the squash commit: "8 cases" (27 `it` blocks now), "42 suites / 341 tests" (373), and "Known, deliberately not in this PR" still says the room-policy rejection is a bare `Error` when this PR adds `NotAllowedForRoom` (7006). `PluginUsageTracker`, `AUDIO_PLUGIN_CALL_TIMEOUT`, `releaseGraph` and the `cleanup()` try/finally are not mentioned at all.
- Comment volume runs against the 1-3 line convention here: 15 lines on `queue`, 12 on `releaseGraph`, 11 on the `callPlugin` TODO, 9 on `pluginAddInProgress`. Keep the invariants, move the rationale to the body.

packages/hms-video-store/src/media/tracks/HMSLocalAudioTrack.ts
raviteja83: **Retracting this — the premise does not hold in production.**

I checked `client.device.list.success` in `default.kafka_ingest`, 24h of web SDK traffic, 391,750 device listings:

| | count |
|---|---|
| listings with at least one `audioInput` | 269,308 |
| ...with **any** blank label | **0** |
| ...blank label alongside a real (hashed) `deviceId` | **0** |
| listings with no `audioInput` at all | 122,442 |

So a permissioned mic always carries a label, and denied or ungranted permission produces an **empty** input list rather than blank-labelled entries. The `label === ''` disjunct is there for the denied path — specifically the empty track Firefox builds from the audio context, which is what the function's own comment says — and not for real microphones. There is no browser in our traffic handing back a live mic with no label.

My probe constructed that track by hand, so it proved the branch behaves that way, not that the input ever arrives. That is on me: I should have measured the input before calling it a regression.

No change needed here, and the `WeakSet` suggestion is not worth the churn. The other four stand.

packages/hms-video-store/src/plugins/audio/AudioPluginsAnalytics.ts
```diff
@@ -44,6 +44,12 @@ export class AudioPluginsAnalytics {
     }
   }
 
+  cleanup() {
+    for (const name of Object.keys(this.pluginAdded)) {
```
raviteja83:
```suggestion
    for (const name  in this.pluginAdded) {
```

raviteja83 (approved):

PR #3756, opened 2026-09-10
fix(store): recover the first subscribe request when the SFU drops it

raviteja83:
The SFU creates the api data channel, so by DCEP the client's `open` fires a half-RTT before the SFU's, and the first `prefer-audio-track-state` / `prefer-video-track-state` of a session can be discarded before it arrives. Two things then kept it lost: the reply timeout is 10s even though the SFU answers in single-digit ms, and `audio`/`video` hold the desired state that repeat requests dedupe against, so a request the SFU never applied silently swallowed every later attempt to reach that same state.

A request on a channel that has not yet been answered is now retried after 500ms instead of 10s, and subscription state is deduped only against a value the SFU acknowledged or one already on the wire.

Seen in production as a remote peer silent for a whole session, or a recording with no video at all.

### DIFF
--- a/packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
+++ b/packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
@@ -9,13 +9,16 @@ export class HMSRemoteStream extends HMSMediaStream {
   private audio = true;
   private video = HMSSimulcastLayer.NONE;
   /**
-   * `audio` and `video` hold the state we want; these say whether the SFU acknowledged it. Repeat
-   * requests dedupe against the desired value, so without this a request the SFU never applied -
-   * the first of a session is the one at risk - would silently swallow every later attempt to
-   * reach that same state, and the track never recovers.
+   * `audio` and `video` hold the state we want; these two pairs say how far it got. Repeat requests
+   * dedupe against the desired value, so without them a request the SFU never applied - the first
+   * of a session is the one at risk - silently swallows every later attempt to reach that same
+   * state and the track never recovers. Deduping needs both: `confirmed` alone would re-send on
+   * every resize while a request is still in flight.
    */
-  private audioConfirmed = true;
-  private videoConfirmed = true;
+  private confirmedAudio = true;
+  private confirmedVideo = HMSSimulcastLayer.NONE;
+  private inFlightAudio?: boolean;
+  private inFlightVideo?: HMSSimulcastLayer;
 
   constructor(nativeStream: MediaStream, connection: HMSSubscribeConnection) {
     super(nativeStream);
@@ -23,29 +26,40 @@ export class HMSRemoteStream extends HMSMediaStream {
   }
 
   async setAudio(enabled: boolean, trackId: string, identifier?: string) {
-    if (this.audio === enabled && this.audioConfirmed) {
+    if (this.isAudioSettled(enabled)) {
       return;
     }
 
     this.audio = enabled;
-    this.audioConfirmed = false;
+    this.inFlightAudio = enabled;
     HMSLogger.d(
       `[Remote stream] ${identifier || ''}
     streamId=${this.id}
     trackId=${trackId}
     subscribing audio - ${this.audio}`,
     );
-    await this.connection.sendOverApiDataChannelWithResponse({
-      params: {
-        subscribed: this.audio,
-        track_id: trackId,
-      },
-      method: 'prefer-audio-track-state',
-    });
-    // a newer request owns the field by now and will confirm its own value
-    if (this.audio === enabled) {
-      this.audioConfirmed = true;
+    try {
+      const response = await this.connection.sendOverApiDataChannelWithResponse({
+        params: {
+          subscribed: this.audio,
+          track_id: trackId,
+        },
+        method: 'prefer-audio-track-state',
+      });
+      // a dropped response never reached the SFU; the request that replaced this one owns the state
+      if (!response?.dropped) {
+        this.confirmedAudio = enabled;
+      }
+    } finally {
+      if (this.inFlightAudio === enabled) {
+        this.inFlightAudio = undefined;
+      }
     }
+  }
+
+  /** true when the SFU is known to be on `enabled`, or a request for it is already on the wire */
+  private isAudioSettled(enabled: boolean) {
+    return this.inFlightAudio === undefined ? this.confirmedAudio === enabled : this.inFlightAudio === enabled;
   }
 
   /**
@@ -54,6 +68,12 @@ export class HMSRemoteStream extends HMSMediaStream {
    * @param layer is simulcast layer to be set
    * @param identifier is stream identifier to be printed in logs
    */
+  /** the SFU telling us where it already is - authoritative, so no request is needed to reach it */
+  setVideoLayerFromServer(layer: HMSSimulcastLayer, identifier: string, source: string) {
+    this.confirmedVideo = layer;
+    this.setVideoLayerLocally(layer, identifier, source);
+  }
+
   setVideoLayerLocally(layer: HMSSimulcastLayer, identifier: string, source: string) {
     this.video = layer;
     HMSLogger.d(`[Remote stream] ${identifier}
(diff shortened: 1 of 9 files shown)


cyril-k-031225 (commented):
permission probe

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
```diff
@@ -41,6 +68,12 @@ export class HMSRemoteStream extends HMSMediaStream {
    * @param layer is simulcast layer to be set
    * @param identifier is stream identifier to be printed in logs
    */
+  /** the SFU telling us where it already is - authoritative, so no request is needed to reach it */
+  setVideoLayerFromServer(layer: HMSSimulcastLayer, identifier: string, source: string) {
+    this.confirmedVideo = layer;
+    this.setVideoLayerLocally(layer, identifier, source);
```
cyril-k-031225, line 89: **FIX:** `setVideoLayerFromServer` sets `confirmedVideo` / local layer but does not clear `inFlightVideo`.

While a client `setVideoLayer` is in flight, `isVideoLayerSettled` still keys off the in-flight layer, so the SFU-reported state is ignored for dedupe. When that in-flight request later succeeds, it can also overwrite this server-authoritative `confirmedVideo`.

Clear or invalidate `inFlightVideo` here (and only confirm in the client path if the desired layer is still that request’s layer).

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
```diff
@@ -63,13 +96,37 @@ export class HMSRemoteStream extends HMSMediaStream {
...
+          max_spatial_layer: this.video,
+          track_id: trackId,
+        },
+        method: 'prefer-video-track-state',
+      })
+      .then(response => {
+        // a dropped response never reached the SFU; the request that replaced this one owns the layer
+        if (!response?.dropped) {
+          this.confirmedVideo = layer;
+        }
```
cyril-k-031225: **FIX:** Confirming `confirmedVideo = layer` on any non-`dropped` response can clobber a newer desired/confirmed state if an older request settles later.

Guard with something like `if (!response?.dropped && this.video === layer)` (same pattern needed in `setAudio` for `confirmedAudio = enabled`).

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
```diff
@@ -8,31 +8,58 @@ export class HMSRemoteStream extends HMSMediaStream {
...
+        params: {
+          subscribed: this.audio,
+          track_id: trackId,
+        },
+        method: 'prefer-audio-track-state',
+      });
+      // a dropped response never reached the SFU; the request that replaced this one owns the state
+      if (!response?.dropped) {
+        this.confirmedAudio = enabled;
+      }
```
cyril-k-031225, line 61: **FIX (sibling of video):** Same stale-confirm risk — `confirmedAudio = enabled` uses the call’s captured `enabled`, not “desired is still this value”. Prefer `this.audio === enabled` (and/or only confirm when this call still owns `inFlightAudio`).

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
```diff
@@ -8,31 +8,58 @@ export class HMSRemoteStream extends HMSMediaStream {
...
   private video = HMSSimulcastLayer.NONE;
+  /**
+   * `audio` and `video` hold the state we want; these two pairs say how far it got. Repeat requests
+   * dedupe against the desired value, so without them a request the SFU never applied - the first
+   * of a session is the one at risk - silently swallows every later attempt to reach that same
+   * state and the track never recovers. Deduping needs both: `confirmed` alone would re-send on
+   * every resize while a request is still in flight.
+   */
+  private confirmedAudio = true;
+  private confirmedVideo = HMSSimulcastLayer.NONE;
```
cyril-k-031225, line 20: **CONSIDER:** `confirmedAudio = true` at construction matches pre-PR dedupe (first `setAudio(true)` was already a no-op via `this.audio === enabled`). If the SFU can start out of sync with that default, leave confirmed unset/`undefined` so the first sync actually sends.

cyril-k-031225 (changes requested):
## PR review: 100mslive/web-sdks#3756 — fix(store): recover the first subscribe request when the SFU drops it

**Verdict:** request changes
**Summary:** The confirmed/in-flight split and `dropped` flag are the right shape for the silent-swallow bug, and the unproven-channel 500ms retry is well motivated. Two confirmation edges still let an in-flight client request fight a server-authoritative update (and a stale success can overwrite a newer confirm).

### FIX
- `HMSRemoteStream.setVideoLayerFromServer` — does not clear `inFlightVideo`; a later success from the in-flight request can overwrite the SFU-reported layer
- `HMSRemoteStream.setVideoLayer` / `setAudio` confirm paths — confirm the call’s local `layer`/`enabled` without checking that desired state is still that value

### CONSIDER
- Initial `confirmedAudio = true` matches pre-PR dedupe (first `setAudio(true)` was already a no-op); if the SFU can start out of sync with that default, starting `confirmedAudio` unset would force a real sync

### Notes
- CI: `build` success; Vercel success
- Sibling hunt: audio and video share the same confirm/in-flight pattern
- Tests cover failure/retry/dropped well; missing coverage for “server layer update while client request in flight”

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
raviteja83: Fixed. `setVideoLayerFromServer` now drops the in-flight claim as well as writing `confirmedVideo`, so the SFU's own value reaches the dedupe instead of losing to a request still on the wire. Test: `lets a layer pushed by the SFU settle the dedupe while a request is in flight`, validated by reverting the fix.

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
raviteja83: Fixed, taking your guard. It is genuinely reachable: `sendMessage` returns a response the SFU actually sent regardless of the claim (the `if (response) return response` path), so a replaced request that gets a real success — easy once its first attempt has been retried — can be answered *after* the one that replaced it. The confirm is now gated on the desired value still being this one, folded with the dropped/error check into `isApplied(response, stillDesired)` shared by both paths. Test: `does not let a late reply for a replaced request confirm its layer`.

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
raviteja83: Fixed the same way — `isApplied(response, this.audio === enabled)`. The dedupe guard itself had the same shape of bug and is now `this.audio === enabled && this.isAudioSettled(enabled)`, mirroring video's `currLayer === targetLayer && isVideoLayerSettled(...)`; without the desired-state term, a request the SFU applied but that failed locally made the correcting call a silent no-op.

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
raviteja83: Keeping `true`, and I think it is load-bearing. Remote audio is subscribed at the SFU by default — that is why the `setVolume` round trip is only an optimisation, and why no `subscribed: true` is ever sent at join today. Leaving it `undefined` would make the first `setAudio(true)` send a redundant request for every remote audio track at join, i.e. more traffic on the api channel in exactly the window whose open race causes this bug class in the first place.

cyril-k-031225 (approved):
## PR review: 100mslive/web-sdks#3756 — fix(store): recover the first subscribe request when the SFU drops it

**Score:** 9.4/10
**Verdict:** approve
**Summary:** Prior FIX items are addressed. `setVideoLayerFromServer` now clears the in-flight claim, and both audio/video confirm paths gate on `isApplied(..., stillDesired)`. Regression tests cover the SFU-push-while-in-flight and late-replaced-reply cases. The unproven-channel 500ms retry + `channelProven` split remains solid.

### FIX
- none

### CONSIDER
- none new — keeping `confirmedAudio = true` is load-bearing for default-subscribed remote audio (agreed with your reply; forcing an unset confirm would spam the api channel in the open-race window)

### Notes
- Re-reviewed head `daca7de0` after re-request
- Prior threads (inFlight clear, stale confirm on audio/video) verified fixed in code + tests
- CI: `build` success on this head
- Sibling hunt: audio/video share `isApplied` / seq-stamped request claims consistently
- Score >9 with no FIX → approve

PR #3757, opened 2026-09-11
fix(store): report subscribe requests the SFU does not answer

raviteja83:
A subscribe request the SFU never receives has no signal anywhere — no SDK event, no SFU metric, and the browser console only reaches us inside a beam's uploaded Chrome log — so we cannot say how often it happens.

This adds `subscribeRequestRetry` when an attempt goes unanswered and `subscribeRequestUnanswered` when none of them are. It also stops an error reply being recorded as state the SFU applied, and keeps the desired audio state in step when a request is deduped away.


packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
```diff
@@ -263,13 +286,20 @@ export default class HMSSubscribeConnection extends HMSConnection {
...
         lastAttemptError = error as Error;
         HMSLogger.w(this.TAG, `Attempt failed for ${requestId}`, { request, try: i + 1, error });
         if (this.closed) {
           break;
         }
+        // the only signal that this class of miss happened at all - nothing else records it
+        this.publishRequestEvent(AnalyticsEventFactory.subscribeRequestRetry, stateKey, {
+          attempt: i,
+          unproven: i === 0 && !this.channelProven,
+        });
```
cyril-k-031225: **FIX:** `unproven: i === 0 && !this.channelProven` diverges from `waitForResponse`'s `firstSend && !this.channelProven`.

If attempt 0 fails on `waitForChannelOpen` without ever calling `send`, attempt 1 is still the first wire send and still gets the 500ms unproven bound — but a timeout there records `unproven: false` because `i === 1`. That mislabels the channel-open race these events exist to measure.

Capture the same condition used for the timeout (e.g. stash `firstSend && !this.channelProven` before `waitForResponse`, and for open-wait failures use `!sentOnce && !this.channelProven`), and assert `unproven: true` on that path in `subscribeRequestTelemetry.test.ts`.

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
```diff
@@ -8,31 +9,84 @@ export class HMSRemoteStream extends HMSMediaStream {
...
+    return stillDesired && !response?.dropped && !response?.error;
+  }
+
+  /** the SFU telling us where it already is - authoritative, so no request is needed to reach it */
+  setVideoLayerFromServer(layer: HMSSimulcastLayer, identifier: string, source: string) {
+    this.confirmedVideo = layer;
+    // drop any claim still on the wire: it describes a layer the SFU has just contradicted, and
+    // isVideoLayerSettled prefers the claim, which would hide this value from the dedupe
+    this.videoRequest = undefined;
+    this.setVideoLayerLocally(layer, identifier, source);
```
cyril-k-031225, line 91: **CONSIDER:** Clearing `videoRequest` here correctly unblocks dedupe against the SFU-pushed layer, but the in-flight prefer-layer request still holds `latestRequestPerState` on the connection and can keep retrying. Local `confirmedVideo` will not flip on a late success (`stillDesired` fails), yet the SFU may still apply those older bytes after degradation. If that race matters, supersede/cancel at the connection when the server pushes a layer.

cyril-k-031225 (changes requested):
## PR review: 100mslive/web-sdks#3757 — fix(store): report subscribe requests the SFU does not answer

**Score:** 8.0/10
**Verdict:** request changes
**Summary:** Solid fix for silent unanswered subscribe requests: confirmed vs in-flight state on `HMSRemoteStream`, short first-send timeout on an unproven API channel, `dropped` so local resolutions are not treated as SFU acks, and good regression coverage. One FIX on the new telemetry: `unproven` does not match the timeout logic, so the channel-open race this PR exists to measure will be mislabeled.

### FIX
- `subscribeConnection.ts` (`subscribeRequestRetry`) — `unproven: i === 0 && !this.channelProven` does not match `waitForResponse`'s `firstSend && !this.channelProven`. If attempt 0 dies on the channel-open wait without sending, the first real send is attempt 1 (still on the 500ms bound) but a timeout there reports `unproven: false`. That undercounts the open-race signal the events were added for. Align the flag with the same `firstSend` / `sentOnce` condition used for the timeout (and cover it in `subscribeRequestTelemetry.test.ts`).

### CONSIDER
- `HMSRemoteStream.setVideoLayerFromServer` — clears the local `videoRequest` claim and confirms the SFU layer, but does not supersede the in-flight prefer-layer claim in `latestRequestPerState`. A late success for the old request still will not update `confirmedVideo` (good), yet the SFU may have applied those bytes after pushing degradation. Worth a connection-level supersede or cancel if that race is in scope.
- `subscribeRequestRetry` is published in `catch` even when the next loop iteration immediately returns `dropped()` for supersession — minor noise on the retry rate.

### Notes
- CI: `build` success. No prior human reviews.
- Sibling hunt: timeout path correctly uses `sentOnce`; only the analytics property diverges. Call sites of `sendOverApiDataChannelWithResponse` look fine after dropping the unused `requestId` param.
- Score 8.0: strong design and tests; mild FIX on the PR's primary analytics contract.

packages/hms-video-store/src/connection/subscribe/subscribeConnection.ts
raviteja83: Fixed in 5502b518. The flag now mirrors what `waitForResponse` was actually given: `unprovenWait = firstSend && !this.channelProven`, captured per attempt just before the wait.

One deviation from the suggestion, deliberately. For the open-wait failure I record `unproven: false` rather than `!sentOnce && !this.channelProven` — that attempt never put bytes in front of the SFU, so it raced nothing; labelling it true would trade your undercount for an overcount on the same metric. To keep the two failures distinguishable in the data I added a `sent` property, so a channel that never opened reads as `sent: false, unproven: false` and the real race reads as `sent: true, unproven: true`.

Test: `labels the first send unproven even when an earlier attempt never sent` drives attempt 0 into the open-wait timeout, opens the channel, and asserts attempt 1 carries `unproven: true` while the attempt that never sent carries `false`. It fails on the `i === 0` version.

packages/hms-video-store/src/media/streams/HMSRemoteStream.ts
raviteja83: Real, and I am leaving it open rather than resolving it — I have not fixed it here.

You are right that the claim survives: `latestRequestPerState` still points at the old request, so its retries replay the original bytes. Local state is safe (`isApplied` gates on `this.video === layer`, so a late success cannot flip `confirmedVideo` back), but the SFU can end up with a preferred layer the client no longer wants, and on bandwidth recovery it restores to a layer nobody asked for. Given we have a live layer-flap issue elsewhere, I would rather not hand the allocator a stale preference.

The fix is small in the connection — dropping the `${method}:${track_id}` claim makes `superseded()` true and the in-flight retries resolve as `dropped` — but `setVideoLayerFromServer(layer, identifier, source)` has no `trackId` today, so it needs a signature change and its callers updated. That is a behaviour change in the subscribe path, and this PR is deliberately scoped to instrumentation only, so I would rather do it as its own PR than widen this one. Happy to open it next if you agree.

raviteja83: Fixed in 370cb929 — folding it in here rather than a follow-up.

`HMSSubscribeConnection.cancelPendingRequest(method, trackId)` drops the claim for one `${method}:${track_id}`, so anything still chasing it hits `superseded()` on its next turn and resolves as `dropped` instead of replaying the overruled bytes. `setVideoLayerFromServer` now takes the `trackId` it needs to call it; `HMSRemoteVideoTrack.setLayerFromServer` passes `this.trackId`. The key has one spelling now (`stateKeyFor`) shared by the request path and the cancel.

Test: `stops a request the SFU has already overruled` — request for HIGH in flight, SFU pushes LOW, then run out the full retry window and assert exactly one send and a `dropped` resolution. Before the fix the retries replay and it throws.

I also took your second point from the review body: `subscribeRequestRetry` is no longer published when the request is about to be dropped as superseded. Being replaced is not the SFU missing anything, and counting it inflated the rate this PR exists to measure. Covered by `does not count a retry for a request a newer one has replaced`, validated by reverting the guard.

cyril-k-031225 (approved):
## PR review: 100mslive/web-sdks#3757 — fix(store): report subscribe requests the SFU does not answer

**Score:** 9.4/10
**Verdict:** approve
**Summary:** Re-review of HEAD `370cb929`. Both items from the earlier request-changes pass: `unproven` now tracks the first wire send (with `sent` to keep open-wait misses distinct), and `cancelPendingRequest` stops overruled prefer-layer retries. Tests cover the regressions; build is green.

### FIX
- None

### CONSIDER
- `cancelPendingRequest` only drops the claim — an in-flight `waitForResponse` still runs to timeout before the next `superseded()` check. Fine for correctness (resolves `dropped`); only matters if you ever want faster teardown under heavy degradation flaps.

### Notes
- Prior FIX (`i === 0` vs `firstSend`) addressed in `5502b518` with a deliberate `sent`/`unproven` split; test `labels the first send unproven even when an earlier attempt never sent` locks it.
- Prior CONSIDER (connection claim surviving `setVideoLayerFromServer`) addressed in `370cb929`; test `stops a request the SFU has already overruled` locks it. Superseded retries no longer inflate `subscribeRequestRetry`.
- CI: `build` success on latest head.
- Sibling hunt: `setVideoLayerFromServer` call sites updated; production `HMSSubscribeConnection` construction passes `eventBus`.
- Score 9.4: no FIX; only a mild non-blocking CONSIDER on wait abort timing.

PR #3759, opened 2026-09-17
fix(transport): don't apply a publish answer the peer connection is no longer waiting for

raviteja83:
`performPublishRenegotiation` applied the SFU's answer unconditionally, so when a concurrent renegotiation had already returned the publish connection to `stable`, `setRemoteDescription` threw `SetRemoteDescriptionFailed` (4004) — which is built with `isTerminal = true`, causing `RetryScheduler` to tear down the room. A transient network blip therefore ended the call instead of recovering.

A publish answer is now applied only to the offer it answers, keyed on the publish connection's **object identity** plus a connection-owned `localDescriptionEpoch` bumped on entry to `setLocalDescription`. `signalingState` cannot serve as that key: `setLocalDescription(offer)` is legal in `have-local-offer` as well as `stable` (webrtc-pc 4.3.2), so a racer replaces our pending offer without the state moving, and `handleSFUMigration` swaps `publishConnection` for a different non-null object.

A discard is a counted, non-terminal `PublishAnswerSuperseded` (4008) that the owning retry re-drives — never `resolve(true)`. **The join path is the exception**: there the discard surfaces as the 4004 `SetRemoteDescriptionFailed` apps already handle, which is terminal, because a non-terminal error escaping `transport.join` spins `internalLeave`'s join-in-progress wait forever and `initAfterJoin` never ran. Applied at all three publish-side `setRemoteDescription` sites, not one.

There is **no re-offer**: that was `69f0e9526` and `50dd0dd8a` removed it. It dropped `retryPublishIceFailedTask`'s `iceRestart` — the only constraint any caller passes — and then reported success, so `RetryScheduler` fired `onReconnected()` over a still-failed ICE transport.

## Slot ownership

`RENEGOTIATION_CALLBACK_ID` holds one owner, so whoever takes it must settle the one it displaces — otherwise `publishTrack`/`unpublishTrack` awaits for the rest of the session with the track captured and unpublished. `armRenegotiationCallback` now does that. Three consequences follow, all fixed here:

- **Release at the commit point, not in `finally`.** Holding the slot across `setRemoteDescriptionAndDrainCandidates` let a concurrent `restartICE`/`publishTrack` displace — and therefore reject — a negotiation that then succeeded. It is released once `assertPublishAnswerApplicable` passes; `finally` still covers the earlier exits and the identity gate keeps the newer owner's entry.
- **`publishTrack` reconciles a staged track.** A displaced waiter is a legitimate reject, but `transport.publish` swallows it into analytics and `HMSSdk.addTrack` pushes to `auxiliaryTracks` regardless, so the track went live on the wire with `store.addTrack`/`isPublished` never run — a later `removeTrack` then took `cleanup()` instead of `unpublish` and the SFU was never told. The `isTrackStagedForPublish` reconcile moved from `republishOnMigration` into `publishTrack`, so app `addTrack`, role change and migration all get it.
- **`retryPublishIceFailedTask` attaches its handler first.** It armed `p`, then awaited a full round trip before `await p`; anything arming inside that window rejected `p` unhandled, logging `Uncaught (in promise) HMSException 4008` on every displacement.

## Also in this PR

Beyond the staleness key, and previously only in commit messages:

- `RetryScheduler` backs off repeat `PublishIceConnectionFailed` retries — 0s, then 1s/2s/4s/4s (capped) + jitter — so a losing renegotiation cannot spin OFFERs at RTT speed for the whole 60s budget. **This changes retry cadence for every publish-ICE failure, not just the racing case, and will move the 4005/4006 counts the verification section compares against.**
- `signal.trackUpdate(this.trackStates)` moved into a `finally` on the reconnect path, so mutes made while disconnected reach biz even if the renegotiation threw.
- The publish-side candidate drain is first-answer-gated and now also runs in `performPublishRenegotiation`, since `negotiateOnFirstPublish`'s discard path returns before its own drain.

**Measured impact** (24h, internal test accounts excluded, deduped by `user_id`): 1,578 events / 1,383 user-sessions / 47 accounts. Only **15.8%** of ejected user-sessions rejoin within 180s and **0%** within 30s — 84% never return. Top accounts by volume: 444, 356, 127, 126, 66, 53.

**Verification.** Every discard emits `client.publishAnswerDiscarded` through a single emit point, so no 4008 lands without a matching row and a `leave()` race can be excluded from the gate. A genuine failure that leaves the track staged is swallowed by `publishTrack`'s reconcile but still emits `publish.failed`, so the 4004/4005 baseline below measures the epoch key rather than the reconcile quietly removing its own evidence. Five reasons: `connection_replaced`, `superseded_offer`, `unexpected_state`, `waiter_displaced`, `connection_gone` — the last two are waiter-level, have no offer in flight, and therefore carry no `epoch`/`signaling_state`. Pass: `reason='superseded_offer'` user-sessions ≈ the pre-deploy stale-4004 count on the same accounts; 4005/4006 within each account's own 7-day band **after allowing for the backoff change above**. The **m-line-mismatch 4004 (~100/day, ~22 accounts) is expected to fall too** — it is the same race, not a separate mechanism. The message is always `The order of m-lines in answer doesn't match order in offer`, i.e. the staged local offer is not the one that answer replies to, which is exactly what the epoch key tests; pion derives an answer's m-lines from the offer it applied, so the SFU reordering them would be a spec violation rather than a 22-account steady state. This is a revision: an earlier draft of this description said m-line must not move, on the RCA's observation that only 30.2% of those sessions carry a 1003 versus 67.2% for the stable cohort. That rules out the reconnect route, not Route A (`onRenegotiationNeeded` racing `retryPublishIceFailedTask`), which needs no network event — a Route-A-dominated m-line cohort is precisely what the guard catches. What genuinely stays open is the subscribe path (1 event/3d).

**Negative space.** Not flipping `ErrorFactory.SetRemoteDescriptionFailed`'s hardcoded `isTerminal=true` (the subscribe path never reads it). Not fixing subscribe-side 4004 (~7/day). Not re-arming the `setSFUNodeId` branch-B guard — it is dead after the first migration, but re-arming trades "apply an old offer" for "silently drop an offer". Not de-retrying OFFER in `JsonRpcSignal.call`. Not landing single-flight serialisation, which rewrites 100% of web publish negotiation for a 0.53% race. Not calling `onStateChange(Failed)` when an SFU migration genuinely fails — that is `internalLeave`, and the frequency is unmeasured until the `sfuMigrationIncomplete` this adds has run; a transient `WebSocketConnectionLost` mid-migration would eject a peer the signal reconnect could have carried. Not touching the subscribe-side candidate drain — `HMSSubscribeConnection` clears its list on add, so it has no stranded batch. Left as judgement calls after review: `handleFirstPublishNegotiationError`'s one-caller extraction (it keeps `negotiateOnFirstPublish` under the `complexity: 6` lint ceiling), `handleSFUMigration` as a `time`/`timeEnd` wrapper, and `isSupersededAnswer` staying module-level.


raviteja83:
## Rewritten — the previous two commits were not safe to ship

Third commit replaces both. Full detail below; the short version is that the re-offer traded a counted ejection for silent publish death, and the `signalingState` guard it sat on cannot actually prove an answer belongs to our offer.

### What was wrong

Four probes run against the branch, not arguments:

| | measured |
|---|---|
| `createOffer` calls on the re-offer path | `[{"iceRestart":true}, null]` — the constraint is dropped |
| a renegotiation that applied nothing | `rejectedWith: undefined, resolvedWith: true, appliedAnswers: []` |
| guard vs. a racer holding the pending offer | applied `"our-stale-answer"` — the guard **passed** |
| `signal.offer` throwing in the re-offer | `state: 'have-local-offer'` — PC wedged, no rollback exists in the SDK |

`retryPublishIceFailedTask:1289` is the only caller that passes constraints, and it passes exactly `{iceRestart: connectionState !== 'connected'}` — the ICE restart *is* the repair. The re-offer dropped it, then `resolve(true)` made `RetryScheduler` delete the category and fire `onReconnected()` over a still-failed transport. `onConnectionStateChange` fires on transition, so nothing re-arms.

### The key

`signalingState === 'have-local-offer'` proves *a* pending offer exists, never *whose*. `setLocalDescription(offer)` is legal in `have-local-offer` as well as `stable` (webrtc-pc 4.3.2), so a racer replaces ours without the state moving — the `it.todo` the old test file conceded. And `handleSFUMigration` (un-awaited from `setSFUNodeId:443`) leaves `publishConnection` non-null but holding a *different* object, which no null check sees.

Now: publish-connection object identity + a connection-owned `localDescriptionEpoch`, bumped **on entry** to `setLocalDescription`. On entry is load-bearing — the native call is chained onto the operations queue synchronously, so a racer that has merely *entered* SLD is already ahead of our SRD. `HMSConnection.ts` is the only reach to `nativeConnection.setLocalDescription`, so no future call site can forget it.

Applied at all three publish-side sites (`negotiateJoinWebRTC`, `negotiateOnFirstPublish`, `performPublishRenegotiation`), not one. `negotiateOnFirstPublish` is the audience→host path GUVI and Entri run constantly.

### Discard policy

Counted, non-terminal `PublishAnswerSuperseded` (4008) — never `resolve(true)`. `RetryScheduler` re-drives with a fresh `createOffer` and a freshly-decided `iceRestart`. **Terminal on the join path only**: a non-terminal error escaping `transport.join` spins `internalLeave`'s `isJoinInProgress` wait at 10Hz forever, because join's `finally` cannot run from inside that catch. That one would have been worse than the bug.

New code, not 4004 — `publish.failed`/4004 keeps meaning "we applied an answer the browser rejected", so it stays a live residual-hole detector.

### Also fixed, each reachable from the new throw

- `handleSFUMigration` republished in sequence with no error handling: one failure left the remaining tracks unpublished with their local tracks already stopped, as an unhandled rejection.
- `retrySignalDisconnectTask` skipped `signal.trackUpdate` on a throw, losing mutes made during the disconnect.
- `performPublishRenegotiation` returned without settling when the connection was gone — `publishTrack` awaited for the session.
- `PublishIceConnectionFailed` retried at **0ms**; repeats now back off so a lost race cannot spin OFFERs at RTT speed for the whole 60s budget.

### Negative space

Not touching `ErrorFactory.SetRemoteDescriptionFailed`'s hardcoded `isTerminal=true` — the subscribe path at `transport/index.ts:177` never reads `isTerminal`, so flipping it globally loses signal and gains nothing. Not fixing subscribe-side 4004 (~7/day). Not re-arming the `setSFUNodeId` branch-B guard: it is dead after the first migration, but re-arming trades "apply an old offer" for "drop an offer" = silent missing remote media. Not de-retrying OFFER in `JsonRpcSignal.call` — real, but it needs biz and is its own PR. Not landing single-flight serialisation: it rewrites 100% of web publish negotiation to fix a 0.53% race.

The m-line-mismatch 4004 (97 events/24h, 11 accounts) is a **different mechanism**, stays terminal, and must not move in verification.

### Tests

`publishRenegotiationRacedByReconnect.test.ts` deleted, not patched — three of its five tests asserted `resolvedWith === true` on paths where `appliedAnswers` was empty, i.e. the silent-success defect written down as the contract.

New `publishAnswerStaleness.test.ts`: 13 tests on the real `HMSTransport` via `test/helpers/makeTransport`, with a fake native PC that models the W3C operations chain so "a racer has entered SLD but it has not resolved" is expressible. Each of the 9 load-bearing changes was reverted individually and confirmed red. Suite: 44 files / 413 tests, tsc and eslint clean.

cyril-k-031225 (dismissed):

packages/hms-video-store/src/transport/index.ts
```diff
@@ -624,19 +638,24 @@ export default class HMSTransport {
...
     const stream = track.stream as HMSLocalStream;
     stream.removeSender(track);
-    await p;
-    await track.cleanup();
-    if (track.source === 'screen' && this.screenStream) {
-      // stop older screenshare tracks to remove the screenshare banner
-      this.screenStream.forEach(stream => {
-        stream.getTracks().forEach(_track => {
-          _track.stop();
+    try {
```
sapta100ms, line 656: Teardown now runs in the `finally` but the rejection still propagates, and no caller handles "torn down **and** failed". `HMSSdk.removeTrack` (sdk/index.ts:1046) aborts before the `auxiliaryTracks.splice()` and `TRACK_REMOVED`; `RoleChangeManager` does the same before clearing `localPeer.videoTrack`.

Stop a screenshare while a renegotiation loses the epoch race → media is stopped and gone from the store, but `localPeer` still references it and the app never gets `TRACK_REMOVED`. Frozen tile that can't be dismissed, role change half-applied.

Either swallow the error after teardown, or move the teardown into the callers' cleanup.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -624,19 +638,24 @@ export default class HMSTransport {
...
+      await p;
+    } finally {
+      // the sender is already detached and the removal rides the next offer, so local
+      // teardown must not wait on the answer — a reject here would leave the camera live
+      await track.cleanup();
```
sapta100ms, line 675: If `track.cleanup()` rejects (plugin/processor teardown), it replaces the `PublishAnswerSuperseded` rejection *and* skips the rest of the block — the `screenStream` stop loop and `store.removeTrack`. That leaves the screenshare banner up, which is what this `finally` was added to prevent. Worth its own try/catch.

sapta100ms (commented):
The core fix (epoch + connection-identity guard, discard kept non-terminal off the join path) looks right. Four notes on the collateral changes inline.

Also: the description says a re-offer replaces the discarded answer — that was in `69f0e9526` but `50dd0dd8a` removed it, so the shipped code only discards and rejects. Worth correcting before merge.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,16 +1024,18 @@ export default class HMSTransport {
...
     try {
-      const offer = await this.publishConnection.createOffer(this.trackStates);
-      await this.publishConnection.setLocalDescription(offer);
+      const offer = await connection.createOffer(this.trackStates);
+      const epoch = await connection.setLocalDescription(offer);
       const answer = await this.signal.offer(offer, this.trackStates);
-      await this.publishConnection.setRemoteDescription(answer);
-      for (const candidate of this.publishConnection.candidates) {
-        await this.publishConnection.addIceCandidate(candidate);
+      this.assertPublishAnswerApplicable(connection, epoch, HMSAction.PUBLISH);
```
sapta100ms: This throws before `connection.initAfterJoin()` on line 1038, so `onnegotiationneeded` never gets wired on the new publish connection — and every later `publishTrack` then awaits its `RENEGOTIATION_CALLBACK_ID` promise forever.

The discard is non-terminal on the theory that the owner retries, but nothing here does: `handleSFUMigration` is called unawaited from `setSFUNodeId` (line 443) → unhandled rejection, tracks never republished; `handleLocalRoleUpdate` propagates into `HMSSdk.handleLocalRoleUpdate` (sdk/index.ts:1413), which has no catch. The local `catch` only swallows 421, not 4008.

Calling `initAfterJoin()` before the guard would fix the first half.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -444,6 +444,20 @@ export default class HMSTransport {
     }
   }
 
+  /**
+   * Republishes one cloned track during migration. The old track is already stopped by the
+   * time we get here, so one failure must not abort the tracks still to be republished —
+   * the publish ICE retry re-offers all of trackStates anyway.
+   */
+  private async republishOnMigration(track: HMSLocalTrack) {
```
sapta100ms, line 459: `publishTrack` only runs `store.addTrack` / sets `isPublished = true` **after** renegotiation resolves, so swallowing the failure leaves `localPeer.videoTrack` pointing at a track that's absent from the store (no tile, selectors empty) with `isPublished === false` — which later sends `HMSSdk.removeTrack` down the `cleanup()` branch instead of unpublishing, so the SFU keeps receiving it.

The ICE retry re-offer in the comment fixes the SDP, but not this bookkeeping.

raviteja83:
All four confirmed and fixed in `5be8ec62d` — each one was the discard reaching a caller with no handler for it. Also corrected the description: the re-offer is gone, the shipped code only discards and rejects.

**`unpublishTrack` (both notes).** Right on both counts. A superseded answer is now swallowed *after* teardown — the sender is already detached and the removal rides the next offer, so the unpublish did happen and callers must be allowed to finish `splice()`/`TRACK_REMOVED`. Anything else still propagates as before. `track.cleanup()` gets its own try/catch so a processor teardown failure can't skip the `screenStream` stop and `store.removeTrack`.

**`negotiateOnFirstPublish`.** This was the worst of the four — verified both callers: `handleLocalRoleUpdate` propagates into `HMSSdk` at sdk/index.ts:1413 with no catch, and `handleSFUMigration` runs unawaited from `setSFUNodeId:443`. It now calls `initAfterJoin()` and returns `false` rather than throwing. I left it off the healthy path rather than moving it before the guard, so the success path stays byte-identical and only the discard path gains the wiring. The discard is still counted by the analytics event, so returning false isn't silent.

**`republishOnMigration`.** Correct — `publishTrack` does `store.addTrack`/`isPublished = true` only after the renegotiation resolves. Reconciled in the catch, gated on `trackStates.has(publishedTrackId)`: that's the real invariant, since it's what the ICE retry re-offers, and it's false after a subsequent migration cleared `trackStates` (where marking published would be wrong).

Tests: 17 in `publishAnswerStaleness.test.ts`, 44 suites / 417 total, tsc and eslint clean. The four fixes were each reverted individually and confirmed red — including a separate revert that keeps the `return false` but drops `initAfterJoin()`, so the wiring is pinned independently of the not-throwing.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `5be8ec62d`. A `PublishAnswerSuperseded` is now swallowed after teardown so callers reach `splice()`/`TRACK_REMOVED`; every other error still propagates as it did before.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `5be8ec62d`. `track.cleanup()` has its own try/catch, so a processor teardown failure no longer replaces the rejection or skips the `screenStream` stop and `store.removeTrack`. Teardown extracted to `teardownUnpublishedTrack`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `5be8ec62d`. Verified both callers — `handleLocalRoleUpdate` propagates into `HMSSdk` at sdk/index.ts:1413 uncaught, `handleSFUMigration` runs unawaited from `setSFUNodeId:443`. It now calls `initAfterJoin()` and returns `false` instead of throwing. Kept off the healthy path so the success path is byte-identical; the discard is still counted by `client.publishAnswerDiscarded`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `5be8ec62d`. Reconciled in the catch — `store.addTrack(track)` and `isPublished = true`, gated on `trackStates.has(track.publishedTrackId)`. That gate is the real invariant: it is exactly what the ICE retry re-offers, and it is false after a later migration cleared `trackStates`, where marking it published would be wrong.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1051,89 @@ export default class HMSTransport {
...
+   */
+  private handleFirstPublishNegotiationError(ex: unknown): boolean {
+    if (!(ex instanceof HMSException)) {
       throw ex;
     }
+    // resolve for now as this might happen during migration
+    if (ex.code === 421) {
+      return true;
+    }
+    if (isSupersededAnswer(ex)) {
```
sapta100ms: `connection_replaced` is now swallowed like `superseded_offer`, so a stale SFU migration keeps running.

All three discard reasons throw the same `PublishAnswerSuperseded`, and `isSupersededAnswer` keys on the error code — so this can't tell them apart and returns `false` for all three. At `50dd0dd8a` (and on main) a `connection_replaced` discard threw and aborted the losing migration. It no longer does, and `handleSFUMigration:494` ignores the return value entirely (`await this.negotiateOnFirstPublish();`).

Migration A is awaiting negotiation → a second `NodeInfo` (`NotificationManager.ts:173` → `setSFUNodeId:439`) starts migration B, which synchronously runs `clearPeerConnections()` + `createPeerConnections()` + `trackStates.clear()` before its own await. A's guard fires `connection_replaced`, A returns `false` and walks its republish loop against **B's** connection, repopulating the `trackStates` B just cleared. B then clones A's just-published tracks and `cleanup()`s them, killing A's live senders.

Duplicate transceivers, phantom `trackStates` entries reported to biz, dead tiles. `connection_replaced` should abort — the new migration owns everything now; only `superseded_offer`/`unexpected_state` should continue. That needs the reason carried on the error (or a distinct error) rather than inferred from the code.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1051,89 @@ export default class HMSTransport {
...
+      this.publishConnection?.initAfterJoin();
```
sapta100ms: Minor, and not fixable here: this path restores `initAfterJoin()` but not the buffered ICE-candidate flush.

On success the line above drains `for (const candidate of connection.candidates) await connection.addIceCandidate(candidate)`; the return-`false` path skips it. `onTrickle` (:186-193) buffers into `connection.candidates` whenever `remoteDescription` is null, and the only two drains in the file are :1020 and :1061 — `performPublishRenegotiation` sets a remote description but never drains.

So after a role change into a publishing role whose first-publish offer is superseded by a concurrent `retryPublishIceFailedTask`, every candidate trickled before that point is stranded; only candidates arriving after the racer's `setRemoteDescription` reach the PC. Publish ICE can stall into the ICE-failed retry.

Can't just add the flush here (no remote description yet) — `performPublishRenegotiation` is where it belongs, after its `setRemoteDescription`.

sapta100ms (commented):
Re-reviewed at `42425916e`. **All four earlier findings are genuinely fixed** — I checked the code, not just the replies:

- **@653** — `unpublishTrack` catches and rethrows only `!isSupersededAnswer(error)`; confirmed `sdk/index.ts:1046` now reaches `auxiliaryTracks.splice()` + `TRACK_REMOVED`, and `RoleChangeManager.ts:81/98` reaches its cleanup.
- **@672** — teardown extracted to `teardownUnpublishedTrack`, `cleanup()` in its own try/catch; the `screenStream` stop loop and `store.removeTrack` always run.
- **@1059** — `handleFirstPublishNegotiationError` wires `initAfterJoin()`; it's a bare `onnegotiationneeded` assignment, so re-invoking is idempotent.
- **@456** — gate is reachable: `publishTrack` sets `publishedTrackId` and `trackStates` before the first `await`.

The 4008 claim in `42425916` holds too — `RetryScheduler.handleTerminalError` passes the original scheduled error, not the task's 4008. Dropping the explicit `isTerminal` assignment isn't a regression either; `HMSException` defaults it to `false`.

Three new items below, one of which the @1059 fix introduced. Only the first is worth blocking on.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -444,6 +448,28 @@ export default class HMSTransport {
...
+    try {
+      await this.publishTrack(track);
+    } catch (error) {
+      HMSLogger.w(TAG, `republish after migration failed for ${track.trackId}`, error);
+      this.eventBus.analytics.publish(AnalyticsEventFactory.publish({ error: error as Error }));
+      // publishTrack skips its bookkeeping when the answer is discarded, but the transceiver is
+      // attached and trackStates still carries the track, so the ICE retry's offer publishes it.
+      // Leaving isPublished false would send a later removeTrack down cleanup() instead of
+      // unpublish, so the SFU would keep receiving it.
+      if (track.publishedTrackId && this.trackStates.has(track.publishedTrackId)) {
```
sapta100ms: The reconcile fires for every error, not just a superseded answer.

This `catch` is untyped, so a genuine failure — `signal.offer` rejecting with `WebSocketConnectionLost`, or a real `SetRemoteDescriptionFailed` — also takes the `store.addTrack(track)` / `isPublished = true` branch, because `trackStates` was populated at the top of `publishTrack` regardless of outcome.

The app then renders a tile and reports the track published while the SFU never received the offer, and `stream.setMaxBitrateAndFramerate(track)` (:583) was skipped, so the sender keeps browser-default caps.

`isSupersededAnswer()` from this same commit is the natural gate.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `a4625fa2f`. You were right that this was introduced by the @1059 fix. The reason is now returned rather than inferred from the code: `publishAnswerDiscardReason` reports it and emits the analytics event, `publishAnswerDiscardError` builds the throwable, and `assertPublishAnswerApplicable` keeps the throwing shape for the two sites that want it. `negotiateOnFirstPublish` branches — `connection_replaced` throws (aborting migration A before it touches B's connection), `superseded_offer`/`unexpected_state` wire `initAfterJoin()` and return `false`. Also added a `.catch` on `setSFUNodeId`'s unawaited `handleSFUMigration()` so the abort logs instead of surfacing as an unhandled rejection. Test pins the throw and asserts `onnegotiationneeded` is NOT wired on the dead connection.

packages/hms-video-store/src/transport/index.ts
raviteja83: Agreed, and filing rather than fixing here — with one detail that makes it larger than it looks. The publish connection never clears `candidates`: `HMSConnection` documents that `HMSPublishConnection` "keeps the complete list of candidates (for ice-connection failed/disconnect) forever", and only the subscribe path does `candidates.length = 0` (:170). So draining in `performPublishRenegotiation` would re-add every candidate accumulated since join, on every renegotiation. It needs a pending-vs-drained split (or an index) before the flush can move there — noted in the commit body so it is not lost.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `a4625fa2f`. Gated on `isSupersededAnswer(error)`. That is the only case where the reconcile is sound — the transceiver is attached and `trackStates` carries the track, so the ICE retry's offer publishes it. A `WebSocketConnectionLost` or a real `SetRemoteDescriptionFailed` now leaves `isPublished` false, as it should. Test asserts `store.addTrack` is not called on a socket failure.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1054,107 @@ export default class HMSTransport {
...
       const answer = await this.signal.offer(offer, this.trackStates);
-      await this.publishConnection.setRemoteDescription(answer);
-      for (const candidate of this.publishConnection.candidates) {
-        await this.publishConnection.addIceCandidate(candidate);
+      const discarded = this.publishAnswerDiscardReason(connection, epoch, HMSAction.PUBLISH);
+      if (discarded === 'connection_replaced') {
+        // A newer migration owns the peer now — it has already cleared trackStates and built
+        // its own connection. Continuing would republish onto it and duplicate its transceivers,
+        // so abort and let the caller unwind.
+        throw this.publishAnswerDiscardError(discarded, epoch, HMSAction.PUBLISH);
```
sapta100ms: Making this throw hits the **other** caller too, and that one regressed.

`transport.handleLocalRoleUpdate:416` does a bare `await this.negotiateOnFirstPublish()` with no catch, and `HMSSdk.handleLocalRoleUpdate` (`sdk/index.ts:1411-1416`) awaits that *before* `roleChangeManager.handleLocalPeerRoleUpdate` and `interactivityCenter.whiteboard.handleLocalRoleUpdate`.

Promote a peer from a non-WebRTC role while an SFU migration lands during the `signal.offer` round trip → `connection_replaced` → throw → `diffRolesAndPublishTracks` never runs. The promoted peer never captures or publishes camera/mic, `onPeerUpdate(ROLE_UPDATED)` never fires, and the migration has no tracks to republish, so nothing recovers it. The rejection is unhandled as well — `HMSInternalEvent.publish` uses `eventEmitter.emit`, which doesn't await the async subscriber.

Before this commit that path returned `false` and continued. Only `handleSFUMigration` wants the abort — the role-update caller should keep return-false, or catch it.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -476,8 +505,8 @@ export default class HMSTransport {
       const newTrack = localPeer.audioTrack.clone(streamMap.get(stream.id)!);
       this.store.removeTrack(localPeer.audioTrack);
       localPeer.audioTrack.cleanup();
-      await this.publishTrack(newTrack);
       localPeer.audioTrack = newTrack;
+      await this.republishOnMigration(newTrack);
```
sapta100ms, line 576: The `connection_replaced` abort only covers the negotiate window; this loop is still unguarded.

Each `republishOnMigration` here is a full offer/answer round trip. If a second `NodeInfo` arrives during one of them, B replaces `publishConnection`, A's in-flight renegotiation rejects with `PublishAnswerSuperseded`, `republishOnMigration` swallows it, and A's loop keeps going — `publishTrack` does `stream.setConnection(this.publishConnection!)`, attaching A's stale clones' transceivers to **B's** connection.

Worse: A's `publishTrack` overwrites the single `RENEGOTIATION_CALLBACK_ID` slot that B's in-flight `publishTrack` is awaiting, and `releaseRenegotiationCallback` is identity-gated, so B's promise never settles and B's `handleSFUMigration` hangs for the rest of the session. A then finishes with `localPeer.auxiliaryTracks = auxTracks`, clobbering B's list.

This is the same duplicate-transceiver / phantom-`trackStates` failure the commit message describes, one step later in the same function. The loop needs the same identity check between tracks — bail when `this.publishConnection !== connection`.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -440,7 +446,30 @@ export default class HMSTransport {
...
+    try {
+      await this.publishTrack(track);
+    } catch (error) {
+      HMSLogger.w(TAG, `republish after migration failed for ${track.trackId}`, error);
+      this.eventBus.analytics.publish(AnalyticsEventFactory.publish({ error: error as Error }));
+      // Only a superseded answer leaves a publishable track behind: the transceiver is attached
+      // and trackStates carries it, so the ICE retry's offer publishes it, and leaving
+      // isPublished false would send a later removeTrack down cleanup() instead of unpublish.
+      // A real failure (offer never reached the SFU) must NOT be reported as published.
+      if (isSupersededAnswer(error) && track.publishedTrackId && this.trackStates.has(track.publishedTrackId)) {
```
sapta100ms: Minor: this gate still infers from the error code, which is the inference the rest of the commit removed as unsound.

`connection_replaced`, `superseded_offer`, `unexpected_state` and the synthetic `PublishAnswerSuperseded(callback.action, 'publish connection is gone')` raised at :1172 all share code 4008. That last one is a case where the track is definitively *not* published — there is no connection — yet it passes the gate.

The other direction: this now excludes `WebSocketConnectionLost` from `signal.offer`, but `publishTrack` has already run `trackStates.set` and `stream.addTransceiver` before the await, and the SFU may have applied the offer before the socket died. The signal-reconnect re-offer of `trackStates` then does publish it, and leaving `isPublished = false` sends a later `removeTrack` down `cleanup()` instead of `unpublish` — the exact bug 5be8ec62 fixed.

Worth keying on the actual state (transceiver attached + `trackStates` carries it) rather than the error code.

sapta100ms (commented):
Re-reviewed `a4625fa2`. **Both claimed fixes are real** — verified in the code:

- **@1086** — `negotiateOnFirstPublish:1064` throws `publishAnswerDiscardError` for `connection_replaced`; `handleFirstPublishNegotiationError` only special-cases 421 so it escapes, and `handleSFUMigration` doesn't catch. The losing migration aborts.
- **@466** — gated on `isSupersededAnswer(error)` at :469 as described.

Typecheck clean; 19 tests in `publishAnswerStaleness.test.ts` and 69 across `src/transport` + `src/connection` pass.

Two new mediums below, one of which this commit introduced, plus two lows. Reply on the ICE-flush thread separately — your premise is right but the blocker is smaller than you think.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -440,7 +446,30 @@ export default class HMSTransport {
       this.subscribeConnection?.setSfuNodeId(id);
     } else if (id && this.sfuNodeId !== id) {
       this.sfuNodeId = id;
-      this.handleSFUMigration();
+      // unawaited by design; a newer migration aborting this one must not surface as a rejection
+      this.handleSFUMigration().catch(error => HMSLogger.w(TAG, 'sfu migration aborted', error));
```
sapta100ms: Minor: this swallows *every* `handleSFUMigration` rejection and labels them all "aborted".

A genuine failure — `signal.offer` rejecting inside `negotiateOnFirstPublish`, or a real `SetRemoteDescriptionFailed` — now produces only a warn line: no analytics event, no `onStateChange`, and the peer is left half-migrated (old connections closed, `trackStates` cleared, no tracks republished, `listener.onSFUMigration` never fired) with nothing to trigger recovery.

Same conflation you just fixed one function below in `republishOnMigration`. Branch on the discard reason and report the rest.

packages/hms-video-store/src/transport/index.ts
sapta100ms: Agreed on deferring, and your premise is right — but the blocker is smaller than stated, so the follow-up should be cheaper than it looks.

Publish really never clears `candidates`: the only `candidates.length = 0` in the repo is the subscribe path at :164, and `HMSConnection.ts:26` documents publish keeping them forever. But `onTrickle:187` only buffers while `connection.remoteDescription` is null — once the first answer is applied, candidates go straight to `addIceCandidate` and never enter the array. So it only ever holds the pre-first-answer batch, not every candidate since join.

That means `if (!connection.remoteDescription) { drain }` in `performPublishRenegotiation` is safe today, without the pending-vs-drained split. Worth noting in the issue so whoever picks it up does not scope the index work unnecessarily.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `d3571065`. `negotiateOnFirstPublish` returns `false` for `connection_replaced` again — it never throws — and `handleSFUMigration` captures `publishConnection` after `createPeerConnections()` and checks identity itself. `initAfterJoin()` still only runs on the same-connection discard; a replaced connection is already closed and B wires its own.

Test `returns false instead of throwing when the connection was replaced under it`; validated by restoring the throw (fails).

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed. `isMigrationSuperseded(connection)` now runs after `negotiateOnFirstPublish` and after every `republishOnMigration` — audio, video, and each iteration of the aux loop. On a bail inside the loop the clones already taken are handed back (`auxiliaryTracks = auxTracks.concat(localPeer.auxiliaryTracks)`); the originals are already `cleanup()`d, so dropping them would lose the screenshare outright.

`handleSFUMigration` is now a thin wrapper so `HMSLogger.timeEnd` survives the early returns.

Tests: `stops the migration republish loop when a newer migration takes the connection`, `hands the already-cloned aux tracks back when the loop aborts`, `aborts the migration itself when a newer one replaced the connection mid-negotiation`. All three fail when the guard is made to always return false.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed — the gate reads state now, not the code. `isTrackStagedForPublish`: `trackStates` names the track **and** `this.publishConnection.hasTransceiver(track.transceiver)` (new on `HMSConnection`, identity against `getTransceivers()`).

Both directions you named are covered: the `:1172` "publish connection is gone" 4008 fails the gate (no live connection owns the transceiver) even though `trackStates` still carries it; a `WebSocketConnectionLost` from `signal.offer` passes it, so the reconnect re-offer publishing it no longer leaves `isPublished=false` and a later `removeTrack` down `cleanup()`.

Tests: `reports a socket-loss republish as published when the transceiver is still staged` and `does not report a republish as published when nothing is staged on the live connection`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed structurally rather than by branching: with the abort above returning instead of rejecting, nothing legitimate rejects `handleSFUMigration` any more, so the conflation is gone and the catch reports everything — `HMSLogger.e` plus `publish.failed` carrying the error code (`reportSFUMigrationFailure`).

One half I deliberately did not do: no `onStateChange(Failed)`. That is `internalLeave`, and how often a migration genuinely fails is precisely what is unmeasured today — this PR exists because an unmeasured terminal path was ejecting people. A transient `WebSocketConnectionLost` from `signal.offer` mid-migration would eject a peer the signal reconnect could have carried. Analytics first, size it, then decide whether Failed is the right recovery. Happy to flip it if you disagree.

Test: `counts a genuinely failed migration instead of logging it away as aborted`.

packages/hms-video-store/src/transport/index.ts
raviteja83: You were right that it is cheap — so I landed it here rather than filing it. `onTrickle` only buffers while `remoteDescription` is null, so the array holds the pre-first-answer batch only and no pending/drained index is needed.

It is folded into `HMSConnection.setRemoteDescriptionAndDrainCandidates` (gate on `!this.remoteDescription`, then apply, then drain) and used at all three publish-side sites, which also removes the two duplicated drain loops. `negotiateOnFirstPublish`'s discard path returns before its own drain, so `performPublishRenegotiation` is the negotiation that inherits the buffer.

Test: `drains the pre-first-answer candidate buffer, exactly once`. Nothing filed.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -518,14 +614,19 @@ export default class HMSTransport {
           newTrack.nativeTrack.addEventListener('ended', this.onScreenshareStop);
         }
         track.cleanup();
-        await this.publishTrack(newTrack);
         auxTracks.push(newTrack);
+        await this.republishOnMigration(newTrack);
+        if (this.isMigrationSuperseded(connection)) {
+          // hand the clones back: the newer migration re-clones from auxiliaryTracks, and the
+          // originals are already cleaned up, so dropping them loses the screenshare for good
+          localPeer.auxiliaryTracks = auxTracks.concat(localPeer.auxiliaryTracks);
```
sapta100ms: The hand-back is only sound while B still has its aux loop ahead of it.

To be clear on the mechanics: `concat` does preserve whatever B rebuilt, so this isn't clobbering B's list — the problem is what gets handed back when B is already past this point.

A shifts screenshare `X` out of `auxiliaryTracks`, clones it to `A-X`, and blocks in `republishOnMigration` on a slow `signal.offer` to the unhealthy old node. B starts, runs its *entire* aux loop against a now-empty `auxiliaryTracks`, publishes nothing for aux, and fires `onSFUMigration()`. A's offer resolves, A detects supersession and writes `[A-X]` back.

`localPeer` now holds a clone whose transceiver is on A's closed connection, which nothing will ever offer again — B is finished and has already announced success. Dead screenshare tile for the rest of the session.

The comment says "the newer migration re-clones from `auxiliaryTracks`", and that's exactly the assumption that breaks here. Handing back only helps if B hasn't passed its loop; otherwise A needs to tear its own clone down so the app at least gets `TRACK_REMOVED`, or the migration needs a generation check that makes B re-process. A different array op won't fix it.

The current test pins the hand-back shape but has no positive control that B's list survives *and* is still live.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -624,8 +725,30 @@ export default class HMSTransport {
...
-    await p;
-    await track.cleanup();
+    try {
+      await p;
+    } catch (error) {
+      // A genuine failure must leave the track intact: the caller aborts before its own
+      // bookkeeping, so tearing down here strands a stopped track on localPeer that the store
+      // no longer has and the app was never told to remove.
+      if (!isSupersededAnswer(error)) {
+        throw error;
```
sapta100ms, line 762: Dropping the `finally` fixes the torn-down-but-not-removed half, but the abort isn't atomic — the rollback is missing.

By the time this `throw error` runs, `trackStates.delete(...)` (:705-717) and `stream.removeSender(track)` (:727) have already executed and are not undone.

User stops a screenshare, `signal.offer` rejects with `WebSocketConnectionLost`. `HMSSdk.removeTrack` aborts before its splice, so the track stays on `localPeer.auxiliaryTracks` with `isPublished === true` and stays in the store — while `trackStates` no longer names it and its sender is detached, so no later offer (ICE retry *or* signal reconnect) ever carries it again. The browser keeps capturing with the sharing banner up, and there's no path back to either published or removed.

Previously the capture at least stopped. The new test asserts `cleanup`/`removeTrack` weren't called but says nothing about the surviving half-state, so it pins the inconsistency rather than catching it.

Restoring the `trackStates` entry — and holding `removeSender` until the renegotiation settles — would make the abort actually atomic.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1124,104 @@ export default class HMSTransport {
...
+        // Report failure, never throw: handleLocalRoleUpdate awaits this bare, and a throw
+        // would skip diffRolesAndPublishTracks and leave a promoted peer that never captures
+        // or publishes. handleSFUMigration checks connection identity itself instead.
+        if (discarded !== PublishAnswerDiscardReason.ConnectionReplaced) {
+          // Same connection, a racer's offer won and it carries current state — but wire
+          // renegotiation first, without initAfterJoin every later publishTrack hangs.
+          // A replaced connection is already closed; its new owner wires its own.
+          connection.initAfterJoin();
+        }
+        return false;
```
sapta100ms, line 1170: Not throwing is right, but returning `false` here lets the role update run against a connection whose `onnegotiationneeded` is still `null`.

`ConnectionReplaced` deliberately skips `initAfterJoin()`, and B only wires its own after a full offer/answer round trip. In between, `HMSSdk.handleLocalRoleUpdate` continues into `diffRolesAndPublishTracks`: `publishTrack` arms `RENEGOTIATION_CALLBACK_ID` (:670) and calls `stream.addTransceiver`, the native `negotiationneeded` fires into B's unwired handler and is dropped. B then completes and its own `republishOnMigration` → `publishTrack` overwrites the callback slot.

Nothing ever settles the displaced entry — `releaseRenegotiationCallback` (:1221) is identity-gated, so it deliberately keeps the newer one, but the older owner's promise is never rejected. The role-change publish awaits forever with the camera live and unpublished.

The general fix is for whoever overwrites `RENEGOTIATION_CALLBACK_ID` to reject the displaced waiter; that single-slot map is the root of both this and the hang in the @509 scenario.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -440,13 +445,94 @@ export default class HMSTransport {
...
+  }
+
+  /**
+   * A newer NodeInfo starts its own migration, which replaces publishConnection and rebuilds
+   * trackStates. Every republish below is a full offer/answer round trip, so ownership has to be
+   * re-checked between them: continuing would attach this migration's clones to the newer
+   * connection, overwrite the single RENEGOTIATION_CALLBACK_ID slot the newer one is awaiting
+   * (hanging it for the rest of the session), and clobber the auxiliaryTracks list it rebuilt.
+   */
+  private isMigrationSuperseded(connection: HMSPublishConnection | null) {
```
sapta100ms, line 523: Minor, but it bears directly on the metric you're deferring the `Failed` decision on.

`isMigrationSuperseded` treats *any* `publishConnection` change as a newer migration, and `leave()` also calls `clearPeerConnections()` and nulls it. An in-flight migration racing a user leave therefore logs "superseded by a newer one" and emits `sfuMigrationIncomplete{reason:'superseded'}`.

Unlike `publishAnswerDiscarded`, the new event carries no `transport_state` (`AnalyticsEventFactory.ts:141`), so those benign leave races can't be filtered out of the numbers you're planning to read. Adding `transport_state: TransportState[this.state]` — and/or short-circuiting on `TransportState.Leaving` — keeps the signal clean.

packages/hms-video-store/src/connection/HMSConnection.ts
```diff
@@ -58,6 +65,11 @@ export default abstract class HMSConnection {
     return this.nativeConnection.addTransceiver(track, init);
   }
 
+  /** True when the transceiver is still attached to this connection. */
+  hasTransceiver(transceiver?: RTCRtpTransceiver): boolean {
```
sapta100ms, line 69: Minor: `getTransceivers()` is unguarded here, unlike its neighbours.

This file already wraps `getSenders()`/`getReceivers()` in try/catch with the comment *"the getSenders and getReceivers is throwing errors in load test"* (:167).

`hasTransceiver` is reached only from inside `republishOnMigration`'s catch block — the handler that exists precisely to stop one failed republish from aborting the rest. If `getTransceivers()` throws there, the exception escapes it, the whole migration unwinds, and it's then counted as a genuine failure. Returning `false` from a try/catch keeps the degradation local.

packages/hms-video-store/src/transport/publishAnswerStaleness.test.ts
```diff
@@ -0,0 +1,1006 @@
...
+    t.signal = makeSignal(async () => answer('ans-1'));
+
+    await expect(t.negotiateOnFirstPublish()).resolves.toBe(true);
+
+    expect(native.applied).toEqual([{ answer: 'ans-1', against: 'offer-1' }]);
+    expect(native.addedCandidates).toEqual([{ candidate: 'role-c1' }]);
+    expect(native.onnegotiationneeded).toEqual(expect.any(Function));
+  });
+
+  it('passes a 421 through as success on the first publish, and rethrows anything else', async () => {
```
sapta100ms, line 1259: Test quality: the 421 path isn't driven end-to-end, and it's the same mutant class this commit targets.

These call `handleFirstPublishNegotiationError` directly. Deleting the `catch (ex) { return this.handleFirstPublishNegotiationError(ex); }` wiring from `negotiateOnFirstPublish` leaves the whole suite green — while in production that would turn every 421 during a migration into an aborted role promotion *and* an aborted migration.

A case where `signal.offer` rejects with a 421 and `negotiateOnFirstPublish()` resolves `true` closes it.

sapta100ms (commented):
Re-reviewed `d3571065`, `047bcbb3`, `babbb53e`. **All five are genuinely fixed** — verified in code, 39/39 tests pass:

- **@1067** — `negotiateOnFirstPublish:1133` returns `false` for every reason and never throws; `performSFUMigration` captures the connection at :554 right after `createPeerConnections()` and checks identity itself. The `isFlagEnabled` stub so `doesRoleNeedWebRTC` doesn't short-circuit is a real non-vacuity guard.
- **@509** — four checkpoints at :557, :571, :586, :619, table-driven. The wrapper split does keep `HMSLogger.timeEnd` alive across early returns *and* throws.
- **@469** — `isTrackStagedForPublish:477` reads state (`publishedTrackId` + `trackStates` + `hasTransceiver`); both legs independently pinned.
- **@450** — `reportSFUMigrationFailure:460` logs at error level and emits `sfuMigrationIncomplete{reason:'failed'}`.
- **@1087** — `setRemoteDescriptionAndDrainCandidates` at all three publish sites; the per-candidate try/catch is a genuine improvement on the old unguarded loops.

**On skipping `onStateChange(Failed)`: agreed, and I'd keep it that way.** Ejecting a peer on a transient `WebSocketConnectionLost` that signal-reconnect would have carried is strictly worse than the current gap, and "measure the rate before choosing the recovery" is the right order given this PR exists because an unmeasured terminal path was ejecting people. One caveat inline at :522 about that metric being noisy as emitted.

Three mediums below, all in the new code.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed at the root, as you suggested — `armRenegotiationCallback` now owns the slot and rejects what it displaces with the non-terminal 4008, at all three arm sites (`publishTrack`, `unpublishTrack`, `retryPublishIceFailedTask`). `releaseRenegotiationCallback` keeps its identity gate; the two are now complements rather than one half.

Blast radius is contained: every displaced owner has a handler for 4008 already — `republishOnMigration` swallows it and lets `isTrackStagedForPublish` decide the bookkeeping, `unpublishTrack` swallows it and tears down, `transport.publish` catches into analytics, and the ICE retry re-drives. A hang becomes a counted retry.

This also closes the migration-B hang in the @509 scenario, which is why it was worth doing here rather than deferring.

Test: `rejects the displaced renegotiation waiter instead of stranding it`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Right — the rollback was missing. `takeTrackState` now returns what it removed and the rethrow path puts it back, so the track is no longer published-everywhere-except-in-the-payload-every-offer-is-built-from.

Not holding `removeSender` until the renegotiation settles: `addTransceiver` is not the inverse of `removeSender`, so re-attaching would move the m-line and trade this for a worse failure. Flagging it rather than half-doing it — say the word if you want that half too and I will do it properly in a follow-up.

The test now asserts the surviving state rather than just the absence of teardown: `expect(t.trackStates.get('v1'))`.

packages/hms-video-store/src/transport/index.ts
raviteja83: You are right, and a different array op does not fix it — so the clones are torn down instead of handed back. `teardownUnpublishedTrack` + `TRACK_REMOVED`, so capture stops, the store drops them and the app can clean up. Correct whichever side of its aux loop B is on, and no generation counter needed.

It does mean an aux track is lost when B has *not* passed its loop, where the hand-back would have saved it. That trade is deliberate: a lost screenshare the app is told about beats an undismissable tile it is not.

Test renamed to `tears the already-cloned aux tracks down when the loop aborts` and asserts cleanup + `store.removeTrack` + `TRACK_REMOVED`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Added `transport_state: TransportState[this.state]` to `sfuMigrationIncomplete`, on both the superseded and failed legs. Did not short-circuit on `Leaving` — a leave race is a real supersession and counting it with a filterable field beats not emitting it, so the gate can exclude it the same way the `publishAnswerDiscarded` one does.

Good catch that this bears on the deferred decision; that metric is the whole basis for it. Assertions now pin the field on both legs.

packages/hms-video-store/src/connection/HMSConnection.ts
raviteja83: Guarded — try/catch returning `false`, matching the `getSenders`/`getReceivers` treatment at :167. You were right that this one matters more than it looks: the only caller is `republishOnMigration`'s catch, so an escape unwound the migration and got it recounted as a genuine failure.

Test: `keeps the migration going when getTransceivers throws under load`. It was the one fix of the six with no covering test until I revert-checked it.

packages/hms-video-store/src/transport/publishAnswerStaleness.test.ts
raviteja83: Fixed — `resolves the first publish when signal.offer rejects with a 421` and `rejects a non-421 signal.offer failure out of the first publish` both go through `negotiateOnFirstPublish`. Deleting the `catch { return this.handleFirstPublishNegotiationError(ex) }` wiring now fails. The direct-call test stays as the unit-level pair.

cyril-k-031225 (commented):
**Two-axis review (standards / spec) at `d8316b77`. Score: 7/10.**

Ran the transport suites locally (55/55 pass) and eslint on the changed files (clean).

**What's solid:** epoch + connection-identity is the right discriminator; all three publish-side `setRemoteDescription` sites are guarded; the discard is non-terminal everywhere except join; every negative-space item in the description is honoured in code; 39 scenario tests with mutation checks.

**Correctness (1, inline at `transport/index.ts:683`):**
- `publishTrack` never got the `isTrackStagedForPublish` reconcile that `republishOnMigration` got. The new displaced-waiter rejection makes the half-published state reachable from `transport.publish` / `HMSSdk.addTrack`. Same defect class as 5be8ec6.

**Spec drift — description vs code (fix the description, not the code):**
- "It stays terminal on the join path only" — code throws 4004 `SetRemoteDescriptionFailed` there, not a terminal 4008. Correct per 42425916; the description is stale.
- "until the `publish.failed` this adds has run" — migration failure emits `sfuMigrationIncomplete`, not `publish.failed`. The verification gate names the wrong event.
- "Every discard emits `client.publishAnswerDiscarded`" — two discard paths don't (inline at `:1255`).
- Present only in commit messages, absent from the description: RetryScheduler backoff for `PublishIceConnectionFailed` (inline), `signal.trackUpdate` moved into a `finally`, and the publish-side candidate drain is now first-answer-gated and also runs in `performPublishRenegotiation` (the description only disclaims the subscribe drain). Roughly 40% of the diff is invisible to someone reading the body alone.

**Standards:**
- PR template: "You must list at least one issue." None linked.
- Nits, all judgement calls: `handleFirstPublishNegotiationError` is a one-caller extraction of a two-line `if`, apparently to stay under `complexity: 6`; `handleSFUMigration` is now only a `time`/`timeEnd` wrapper around `performSFUMigration`; `getDelayForRetryCount`'s `failedRetryCount = 0` default is dead (the sole caller always passes it); `isSupersededAnswer` is the file's only module-level predicate while the other new ones are methods.

Not re-raising the `unpublishTrack` sender-restore residual; it is already acknowledged in-thread.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -565,13 +680,7 @@ export default class HMSTransport {
...
     );
     this.trackStates.set(track.publishedTrackId, new TrackState(track));
-    const p = new Promise<boolean>((resolve, reject) => {
-      this.callbacks.set(RENEGOTIATION_CALLBACK_ID, {
-        promise: { resolve, reject },
-        action: HMSAction.PUBLISH,
-        extra: {},
-      });
-    });
+    const p = this.armRenegotiationCallback(HMSAction.PUBLISH);
```
cyril-k-031225, line 679: `publishTrack` is missing the reconcile `republishOnMigration` got in 5be8ec6 / a4625fa2, and the new displaced-waiter rejection makes it reachable from `transport.publish`.

Scenario: app calls `addTrack(screen)` → this arms PUBLISH, `trackStates.set` ran on the line above, `addTransceiver` follows, then `await p`. Publish ICE fails mid-flight → `retryPublishIceFailedTask` arms RESTART_ICE (:1489) → `armRenegotiationCallback` rejects this waiter with 4008 → `transport.publish` (:423) swallows it into analytics → `HMSSdk.addTrack` (sdk/index.ts:1033) still pushes to `auxiliaryTracks` and fires `TRACK_ADDED`. The retry's offer carries `trackStates`, so the SFU publishes the track and remote peers see it, but `store.addTrack` and `isPublished = true` never ran. A later `removeTrack` takes `track.cleanup()` (sdk/index.ts:1048) instead of `unpublish`, `trackStates` keeps the entry, and the SFU is never told. Two concurrent `addTrack` calls trigger the same thing (second arm displaces the first).

Before this PR that waiter hung; now it half-publishes, which is harder to notice.

Suggest putting the reconcile here so all three callers get it, and dropping the copy in `republishOnMigration`:

```ts
try {
  await p;
} catch (error) {
  if (!isSupersededAnswer(error) || !this.isTrackStagedForPublish(track)) {
    throw error;
  }
  // displaced by a newer offer that already carries this track — treat as published
}
```

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1140,123 @@ export default class HMSTransport {
...
+   * entry, so a displaced owner is otherwise never settled and its caller awaits for the rest
+   * of the session with the track captured but unpublished.
+   */
+  private armRenegotiationCallback(action: HMSAction): Promise<boolean> {
+    const displaced = this.callbacks.get(RENEGOTIATION_CALLBACK_ID);
+    return new Promise<boolean>((resolve, reject) => {
+      this.callbacks.set(RENEGOTIATION_CALLBACK_ID, { promise: { resolve, reject }, action, extra: {} });
+      if (displaced) {
+        HMSLogger.w(TAG, `renegotiation waiter for ${displaced.action} displaced by ${action}`);
+        displaced.promise.reject(
```
cyril-k-031225, line 1277: Description: "Every discard emits `client.publishAnswerDiscarded` with `reason`, `action`, `signaling_state`, `transport_state`." This reject and the `publish connection is gone` reject at :1274 both throw 4008 with no event, and `PublishAnswerDiscardReason` has no value for either. The `superseded_offer` gate is still computable, but these 4008s will land in `publish.failed` with no matching `publishAnswerDiscarded` row, which is what the verification section says to reconcile against. Either add `waiter_displaced` / `connection_gone` reasons and emit here, or narrow the description to the three offer/answer sites.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1239,13 +1486,7 @@ export default class HMSTransport {
...
      */
     if (this.publishConnection) {
-      const p = new Promise<boolean>((resolve, reject) => {
-        this.callbacks.set(RENEGOTIATION_CALLBACK_ID, {
-          promise: { resolve, reject },
-          action: HMSAction.RESTART_ICE,
-          extra: {},
-        });
-      });
+      const p = this.armRenegotiationCallback(HMSAction.RESTART_ICE);
```
cyril-k-031225, line 1518: Minor: `p` is armed here, then `performPublishRenegotiation` is awaited across a network round trip before `await p` attaches a handler. If a `publishTrack` / `unpublishTrack` arms during that window, `armRenegotiationCallback` rejects `p` with no handler attached, so the browser logs `Uncaught (in promise) HMSException 4008` and later fires `rejectionhandled`. Outcome is still correct (the retry re-drives), just noisy on every displacement. Attaching first closes it:

```ts
const p = this.armRenegotiationCallback(HMSAction.RESTART_ICE);
await Promise.all([p, this.performPublishRenegotiation({ iceRestart: ... })]);
```

packages/hms-video-store/src/transport/RetryScheduler.ts
```diff
@@ -201,14 +201,18 @@ export class RetryScheduler {
...
-  private getDelayForRetryCount(category: TFC) {
+  private getDelayForRetryCount(category: TFC, failedRetryCount = 0) {
     const jitter = category === TFC.JoinWSMessageFailed ? Math.random() * 2 : Math.random();
     let delaySeconds = 0;
     if (category === TFC.JoinWSMessageFailed) {
       // linear backoff(2 + jitter for every retry)
       delaySeconds = 2 + jitter;
     } else if (category === TFC.SignalDisconnect) {
       delaySeconds = 1;
+    } else if (category === TFC.PublishIceConnectionFailed && failedRetryCount > 0) {
```
cyril-k-031225, line 212: Not in the PR description (only in 50dd0dd8's message) and independent of the staleness key: this changes retry cadence for every `PublishIceConnectionFailed`, not just the racing case. 1s / 2s / 4s / 4s… instead of immediate means fewer attempts fit the 60s budget. Fine if intended, but it needs a line in the description and its own impact note, since it will move the 4005/4006 numbers the verification section compares against.

raviteja83:
Self-review found one more instance of the slot-ownership class: the `finally` release holds `RENEGOTIATION_CALLBACK_ID` across the drain, so a concurrent `restartICE`/`publishTrack` displaces — and now rejects — a negotiation that then succeeds. `publishTrack` throws before `store.addTrack`, leaving the track live on the wire but absent from the store.

`5cba65115` releases at the commit point (`assertPublishAnswerApplicable`) instead; `finally` still covers earlier exits and the identity gate keeps the newer owner. Test revert-validated 3x and mutation-checked — dropping the identity gate or moving the release back past the drain both kill it.

Swept the rest of the diff for the same class: `negotiateOnFirstPublish` never takes the slot, and the `SUBSCRIBE_ICE_CONNECTION_CALLBACK_ID` sites delete immediately after `get`. One instance, not a class.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `e652bc6c1`. Confirmed the whole chain - `transport.publish:420` swallows into analytics, `HMSSdk.addTrack:1033` pushes to `auxiliaryTracks` regardless, `removeTrack:1046` branches on `isPublished`.

Took your suggestion but keyed the gate on staging alone rather than `isSupersededAnswer(error) || ...`, so the `WebSocketConnectionLost`-leaves-it-staged case falls in the same branch - that was the other half of what `republishOnMigration` was doing. Falling through also picks up `setMaxBitrateAndFramerate`, which the old copy skipped. `republishOnMigration`'s copy is dropped, so there is one gate serving all three callers.

The five gate tests moved with it and now drive `publishTrack` directly; added the app `addTrack` and two-concurrent-`addTrack` scenarios you described. Mutation-checked: swallowing everything kills 4 tests, dropping the `trackStates` term kills 1.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `e652bc6c1` - chose "emit" over "narrow the description". Added `waiter_displaced` and `connection_gone`, both routed through a new single `reportPublishAnswerDiscard` emit point. There are now exactly three 4008 throw sites and three emit sites, so no 4008 lands without a row.

Neither has an offer in flight, so `signaling_state`/`epoch` are now optional on the factory and absent on these two rather than faked. Description updated to name all five reasons and that caveat.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `e652bc6c1`, using your `Promise.all` exactly.

One note for the record: `performPublishRenegotiation` never rejects - it catches everything into the callback and returns - so `Promise.all` cannot change which error surfaces, only when the handler attaches.

The test drives it through a `signal.offer` that crosses a macrotask boundary and asserts against a real `process.on('unhandledRejection')` listener. My first attempt passed immediately because the fake resolved inside one microtask turn, which is exactly why this never showed up locally.

packages/hms-video-store/src/transport/RetryScheduler.ts
raviteja83: Fixed in the description, not the code - agreed it is intended. Added under a new "Also in this PR" section with the explicit ramp (0s, then 1s/2s/4s/4s capped, plus jitter) and a callout that it changes cadence for **every** `PublishIceConnectionFailed`, not just the racing case.

Carried the impact note into the verification section too: the 4005/4006 comparison now reads "within each account's own 7-day band **after allowing for the backoff change above**".

Separately, your nit about `getDelayForRetryCount` is right - `scheduleTask` destructures `failedRetryCount = 0` at line 70, so it is always a number by the call at 152. Default removed.

packages/hms-video-store/src/transport/index.ts
```diff
@@ -565,19 +676,23 @@ export default class HMSTransport {
...
+    try {
+      await p;
+    } catch (error) {
+      // A displaced waiter or a lost socket can leave the track fully staged, and the next offer
+      // (ICE retry, signal reconnect) publishes it. Reporting it unpublished sends a later
+      // removeTrack down cleanup() instead of unpublish, and the SFU is never told.
+      if (!this.isTrackStagedForPublish(track)) {
+        throw error;
+      }
+      HMSLogger.w(TAG, `publishTrack: waiter lost its race, ${track.trackId} is staged for the next offer`, error);
```
cyril-k-031225, line 699: The state-based gate is right, but this branch now swallows every error that leaves the track staged, and the only place that used to turn those into a `publish.failed` row was the caller's catch (`transport.publish:423`, and `republishOnMigration:503` before its reconcile moved here). `performPublishRenegotiation`'s catch emits no analytics of its own, so a genuine `SetRemoteDescriptionFailed` (real m-line mismatch) or `WebSocketConnectionLost` that rejects a publish-owned renegotiation now produces a warn log and nothing else.

That matters for this PR's own gate: the verification section expects the m-line-mismatch 4004 count to fall because of the epoch key. After this commit it also falls because the surviving 4004s stop being reported, and the two are indistinguishable in the data.

A displaced waiter already has its `publishAnswerDiscarded` row, so only the non-4008 case needs the event:

```ts
if (!isSupersededAnswer(error)) {
  this.eventBus.analytics.publish(AnalyticsEventFactory.publish({ error: error as Error }));
}
HMSLogger.w(TAG, `publishTrack: waiter lost its race, ${track.trackId} is staged for the next offer`, error);
```

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
```diff
@@ -133,6 +134,47 @@ export default class AnalyticsEventFactory {
...
+        error,
+      ),
+    });
+  }
+
+  /**
+   * A publish answer was dropped because the offer it answers is no longer staged.
+   * `action` names the owner that must re-drive; `reason` separates the three causes.
+   */
+  /** the two waiter-level discards have no offer in flight, so they carry no epoch/signaling_state */
```
cyril-k-031225: Nit: two stacked JSDoc blocks — only this one attaches to the method, so the paragraph above it is orphaned. Also the line above says "`reason` separates the three causes"; it is five now. Merge into one block.

cyril-k-031225 (commented):
**Re-review at `e652bc6c`. Score: 8/10** (was 7).

Verified locally: transport suites 61/61 pass, eslint clean on changed files.

**All four inline findings from the previous round are genuinely fixed** — checked the code, not just the replies:
- `publishTrack:691` now gates on `isTrackStagedForPublish` and falls through to `store.addTrack` / `setMaxBitrateAndFramerate` / `isPublished = true`; `republishOnMigration`'s copy is gone, so one gate serves `addTrack`, role change and migration. `HMSLocalStream.addTransceiver:25` sets `track.transceiver` before the await, so the gate is live in production, not just under mocks.
- Both waiter-level 4008s now emit through `reportPublishAnswerDiscard` with `waiter_displaced` / `connection_gone`; three throw sites, three emit sites.
- `retryPublishIceFailedTask` attaches via `Promise.all` before the round trip; the test uses a real `unhandledRejection` listener across a `setTimeout` boundary, which is the right shape.
- Description now has "Also in this PR" (backoff with cadence callout, `trackUpdate` finally, publish-side drain), says 4004 on join, and no longer names `publish.failed` as the migration gate.

**`5cba6511` (self-found) is sound.** Releasing at the commit point restores roughly the pre-PR release timing (the old code deleted the slot right after `signal.offer`), `finally` still covers the earlier exits via the identity gate, and a concurrent arm during the drain gets its own undisplaced negotiation whose `setLocalDescription` queues behind the in-flight `setRemoteDescription` on the operations chain. The new test covers it.

**One new finding (inline at `transport/index.ts:694`):** the swallow branch drops the `publish.failed` analytics row for a genuine failure that leaves the track staged. Small fix, but it touches the 4004 baseline the verification section reads.

**Still open from standards:** PR template requires at least one linked issue; none yet.

Nit inline at `AnalyticsEventFactory.ts:167`.

packages/hms-video-store/src/transport/index.ts
raviteja83: Fixed in `a7a36e825`, using your gate.

You are right that it lands on this PR's own verification: the reconcile would have removed the evidence the 4004 comparison reads, so a fall in m-line 4004 could not be attributed to the epoch key rather than to the reporting going quiet. `performPublishRenegotiation` confirmed to emit no analytics of its own, so `transport.publish:423` really was the only reporter.

Two tests, both directions: a `WebSocketConnectionLost` that leaves the track staged now produces exactly one `publish.failed`, and a displaced waiter still produces zero (it already has its `publishAnswerDiscarded` row). Dropping the `isSupersededAnswer` gate kills the second; removing the emit kills the first.

Added a clause to the Verification section so the description states this rather than leaving it implicit.

packages/hms-video-store/src/analytics/AnalyticsEventFactory.ts
raviteja83: Fixed in `a7a36e825` — merged into one block. My earlier edit added a second `/** ... */` directly above the method without noticing the existing one, so the upper block was orphaned and kept its stale "three causes".

The merged block now says five, and folds in the note that the two waiter-level reasons carry no `epoch`/`signaling_state`.

Also found and fixed a stale comment of the same kind while checking for siblings: the doc block above the `getTransceivers throws under load` test still placed the gate in `republishOnMigration`'s catch after it had moved to `publishTrack`.

raviteja83:
On the linked-issue standard: this repo is public, and the work came from prod telemetry rather than a filed report, so there is no issue to link. Treating that template line as not applicable here rather than filing one to satisfy it.

cyril-k-031225 (dismissed):

cyril-k-031225 (approved):

packages/hms-video-store/src/transport/index.ts
```diff
@@ -1002,24 +1151,137 @@ export default class HMSTransport {
...
+   * of the session with the track captured but unpublished.
+   */
+  private armRenegotiationCallback(action: HMSAction): Promise<boolean> {
+    const displaced = this.callbacks.get(RENEGOTIATION_CALLBACK_ID);
+    return new Promise<boolean>((resolve, reject) => {
+      this.callbacks.set(RENEGOTIATION_CALLBACK_ID, { promise: { resolve, reject }, action, extra: {} });
+      if (displaced) {
+        HMSLogger.w(TAG, `renegotiation waiter for ${displaced.action} displaced by ${action}`);
+        this.reportPublishAnswerDiscard(PublishAnswerDiscardReason.WaiterDisplaced, displaced.action);
+        displaced.promise.reject(
```
sapta100ms, line 1277: Non-blocking, and strictly an improvement on what it replaced — but worth knowing where the rejection lands.

Settling the displaced owner converts the hang into a rejection on a path that still has no catch: `HMSSdk.handleLocalRoleUpdate` (`sdk/index.ts:1411`) awaits `transport.handleLocalRoleUpdate` bare, and it is invoked from an event emitter that does not await its async subscriber. So a waiter displaced during a role change now surfaces as an unhandled rejection rather than awaiting forever.

Better than the hang either way, and the 4008 is the non-terminal code the retry machinery already handles — so this is a log-noise/observability point, not a correctness one. If you want it quiet, that caller needs a catch.

sapta100ms (commented):
Re-reviewed `d8316b77`. All six are genuinely fixed — `armRenegotiationCallback` settles the displaced owner, `takeTrackState` makes the unpublish abort recoverable, the aux clones are torn down instead of handed back, both `sfuMigrationIncomplete` emissions carry `transport_state`, and `hasTransceiver` degrades to `false`. Typecheck clean, 43/43 in the staleness file, 93/93 across `src/transport` + `src/connection`.

Agreed on not re-attaching the sender — `addTransceiver` isn'''t the inverse of `removeSender` and would move the m-line, which is worse than the gap it would close. Putting the trackState back was the half that mattered.

One non-blocking note inline. Nothing here needs to hold the PR.

